PR #98, opened 2015-01-05
IGN France: change to move tests about `match_type` in the reverse geocode case

ThomasG77:
The case match_type in 'Street enhanced' where no building number is not working the expected way so keeping a special use case for this purpose is not useful.
We added a new test and made a change in test/geocoders/util.py class to also test address key and not only raw, longitude and latitude.

To see the existing issue that justified this change issue, only apply change in util.py and in test/geocoders/ignfrance.py.
You will that geocoding `Camp des Landes, 41200 VILLEFRANCHE-SUR-CHER` will send back `None le camp des landes, 41200 Villefranche-sur-Cher` whereas it should be `le camp des landes, 41200 Villefranche-sur-Cher`
It comes from the facts:
* that when we convert for XML we set keys to None when not available
* the condition in ignfrance.py `_parse_place` function was supposing that `building` key was never empty (to None) with the case `match_type` 'Street enhanced'

### DIFF
--- a/geopy/geocoders/ignfrance.py
+++ b/geopy/geocoders/ignfrance.py
@@ -516,40 +516,25 @@ class IGNFrance(Geocoder):   # pylint: disable=W0223
         if is_freeform == 'true':
             location = place.get('freeformaddress')
         else:
-            # When classic geocoding
-            if place.get('match_type'):
+            # For parcelle
+            if place.get('numero'):
+                location = place.get('street')
+            else:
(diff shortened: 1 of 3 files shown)


ijl:
@ThomasG77, thanks, merged in 3bb5dd85c1b2fbef4937055acda4fdc86742c7d5.

ijl:
Released in 1.7.1.

PR #101, opened 2015-01-17
Support GeocodeFarm V3 geocoder

bigsassy:
GeocodeFarm has a new version of their geocoder.  This pull request adds a new geocoder, GeocodeFarmV3, that supports the new API.

https://www.geocode.farm

Comes complete with:

* New geocoder that can geocode and reverse
* Full test suite
* Updated documentation (not really tested).  Code still needs `..versionadded` stuff in the code.

As far as I can tell, it's good to go. Let me know if there's anything else that should be done.

### DIFF
--- a/geopy/geocoders/geocodefarm.py
+++ b/geopy/geocoders/geocodefarm.py
@@ -17,12 +17,12 @@ __all__ = ("GeocodeFarm", )
 class GeocodeFarm(Geocoder):
     """
     Geocoder using the GeocodeFarm API. Documentation at:
-        https://www.geocodefarm.com/dashboard/documentation/
+        https://www.geocode.farm/geocoding/free-api-documentation/
     """
 
     def __init__(
             self,
-            api_key,
+            api_key=None,
             format_string=DEFAULT_FORMAT_STRING,
             timeout=DEFAULT_TIMEOUT,
             proxies=None,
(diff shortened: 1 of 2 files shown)


ijl:
@bigsassy, I can't see any change in behavior that would make a consumer care they are using v3 versus v1 of the API. Do you know of any? If not, how about just modifying `GeocodeFarm`?

bigsassy:
There isn't any negative change.  In fact there's only benefits, including SSL calls and getting back many results for a query instead of only the most accurate one (as deemed by geocodefarm).

I didn't update the current `GeocodeFarm` geocoder for the sake of anyone still on version 2 and breaking their code if they update to the latest geopy.

That said, I'm perfectly ok (and in fact would prefer) to just upgrade the current geocoder to use version 3, and maybe mention updating your API keys to V3 in the release notes and optionally the documentation.

Let me know how you'd like me to move forward, @ijl.

ijl:
I didn't need to obtain a new API key for v3 to work myself. Did you have an issue there?

I see upon looking at the response more that this'd introduce an avoidable breaking change in the structure of `Location.raw["ADDRESS"]`. I'm okay with that, though, given that it's an opt-in unsafe property.

Looks like we can make `api_key` default to `None` on `__init__` for unauthenticated users?

And adding a `scheme` argument to the `__init__` would break someone constructing from positional arguments, so I suppose let's keep that off and just default to `https`?

But yes modifying `GeocodeFarm` seems best to me as well.

bigsassy:
Ok, I finally got around to updating this.  It appears the IGNFranceTestCase suite is having issues on Travis CI, but otherwise this pull request should be good to go.  Let me know if there's anything I can do before it gets merged.

ijl:
Merged in e1d3e7a0c31dcb8c6587576edf9c6e7eb85cfb37, thanks.

ijl:
Python 3 builds are differing from 2: https://travis-ci.org/geopy/geopy/builds/57030997 

Will need to look at it later, unless you're familiar.

bigsassy:
Fixed in #118.

ijl:
Released in 1.10.0.

PR #102, opened 2015-01-28
fix: add api key to params for googlev3 reverse geocoding

luke:
Just hit an issue where Google was saying we had gone over the geocoding limit, I debugged the issue and traced it back the api_key param not being sent to Google. This change just adds the key to the params if its set.


ijl:
Merged in 336b03a6a6a600d0f5845ef65ea3e7ba31af32fb, thanks.

ijl:
Released in 1.8.1.

PR #104, opened 2015-01-30
Custom nominatim server

srounet:
Allow one to specify a custom Nominatim server.

- adds **domain** and **scheme** parameter to Nominatim class.
- bump version to **1.8.2**
- does not break existing tests

### DIFF
--- a/geopy/__init__.py
+++ b/geopy/__init__.py
@@ -13,4 +13,4 @@ from geopy.location import Location
 from geopy.geocoders import * # pylint: disable=W0401
 
 
-__version__ = "1.8.1"
+__version__ = "1.8.2"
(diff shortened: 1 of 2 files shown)


ijl:
Would it not be a little surprising if `scheme` is not respected when you specify `api`? Did you want to have the same URL for both forward and reverse geocoding? I think it may be better implemented in the same way that GoogleV3 does: https://github.com/geopy/geopy/blob/master/geopy/geocoders/googlev3.py#L42. What do you think of that `domain` implementation?

srounet:
Didn't see the domain implementation in `GoogleV3`, I'll move my code to implement the same kind of behaviour.
As for `forward` and `reverse` geocoding, providing a unique domain resolve most of the common cases, thus, you can still create a second instance of the geocoder of your choice without specifying domain.

I think specifying domain should act like a `sandbox` where everything goes to the domain.

ijl:
Ok. Please ping me when you add the commit so I can merge, as GitHub doesn't notify me.

ijl:
Merged in f514de942f98bf95daeeeba27980be2cf842794f, thanks.

ijl:
Released in 1.9.0.

PR #105, opened 2015-02-11
Adds culture to reverse

7wonders:
Adds the culture setting to Bings reverse geocoding


ijl:
@7wonders, thanks, can you just add a test exercising it to https://github.com/geopy/geopy/blob/master/test/geocoders/bing.py ?

PR #113, opened 2015-03-17
Replaced 'view_box' with viewbox per OSM documentation

trukvl:
Referenced this document: http://wiki.openstreetmap.org/wiki/Nominatim

This does not work (copied from geopy log message):  http://nominatim.openstreetmap.org/search?q=Chicago+IL&view_box=%5B-180%2C+-90%2C+180%2C+90%5D&format=json

This does work: http://nominatim.openstreetmap.org/search?q=Chicago+IL&viewbox=%5B-180%2C+-90%2C+180%2C+90%5D&format=json

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -37,7 +37,7 @@ class Nominatim(Geocoder):
     def __init__(
             self,
             format_string=DEFAULT_FORMAT_STRING,
-            view_box=(-180, -90, 180, 90),
+            viewbox=(-180, -90, 180, 90),
             country_bias=None,
             timeout=DEFAULT_TIMEOUT,
             proxies=None,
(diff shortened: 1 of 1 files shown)


redwire:
+1
I'd really like to see this merged ASAP.
Please fix as soon as you can.

trukvl:
The checks that failed seem unrelated to the change:

ERROR: test suite for <class 'test.IGNFranceTestCase'>

The output indicated that there is no key for the IGN France API.

kpanic:
Weird, I experienced this bug yesterday, but not today:

```
$ pip freeze|grep geo
geopy==1.9.1

In [1]: from geopy import Nominatim
In [2]: geocoder = Nominatim()
In [3]: geocoder.geocode('Potsdamer Platz')
Out[3]: Location((52.5087317, 13.3770476, 0.0))
```

solumos:
It looks like both cases ("viewbox" and "view_box") are now working.

ijl:
Thanks for the pull request. I was away when this became an issue. Changing the parameter name would be an unnecessary breaking change in Nominatim's init, though, so I've pushed 706655f instead.

ijl:
Released in 1.10.0.

PR #115, opened 2015-03-21
Added possibility to add a custom http request user agent header

sebastianneubauer:
My application that uses geopy was blocked by Nominatim as geopy uses the standard user agent header for requests provided by urllib (Python-urllib/2.7). Nominatim requires to add an identifying user agent header per application in order to be able to track and block requests if needed. 

I now added in this PR a default header in the base geocoder class:
User-Agent: geopy/#version
where #version is the current geopy version (e.g "geopy/1.9.1")

This is now added to every request and every geocoder inheriting from the 'Geocoder' base class.
This should be much better, as it is much easier for services to be able to track which request come from where. Now every service is able to distinguish geopy requests from non geopy requests.

In addition I added the possibility to pass in a custom user agent string to the Nominatim geocoder class:
```python
        geocoder = Nominatim(
            user_agent='my_user_agent/1.0'
        )
        self.assertEqual(geocoder.headers['User-Agent'], 'my_user_agent/1.0')
```
This should help to prevent quite some blockings`and bans from Nominatim, as now they are able to distinguish bad requests from good request based on the user agent headers and only block those, which have e.g a standard user agent like "geopy/1.9.1" or even "Python-urllib/2.7" and spare those requests identifying with a specific user agent like "GoodApp/0.1" as long as all "GoodApp/0.1" requests respect the Nominatim terms of usage.

I also added tests for it and thereby fixed the tests for ignfrance which were not able to pass tests without api key credentials even they should be skipped in that case.

### DIFF
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -45,10 +45,7 @@ DEFAULT_FORMAT_STRING = '%s'
 DEFAULT_SCHEME = 'https'
 DEFAULT_TIMEOUT = 1
 DEFAULT_WKID = 4326
-
-
-def get_default_user_agent():
-    return "geopy/" + gu.get_version()
+DEFAULT_USER_AGENT = "geopy/" + gu.get_version()
 
 
 ERROR_CODE_MAP = {
@@ -91,7 +88,7 @@ class Geocoder(object): # pylint: disable=R0921
             )
         self.proxies = proxies
         self.timeout = timeout
-        self.headers = {'User-Agent': user_agent or get_default_user_agent()}
+        self.headers = {'User-Agent': user_agent or DEFAULT_USER_AGENT}
 
         if self.proxies:
             install_opener(
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -29,7 +29,7 @@ from geopy.exc import (
     GeocoderUnavailable,
     GeocoderParseError,
 )
-from geopy.util import decode_page, get_version
+import geopy.util as gu
 
 
 __all__ = (
@@ -45,7 +45,11 @@ DEFAULT_FORMAT_STRING = '%s'
 DEFAULT_SCHEME = 'https'
 DEFAULT_TIMEOUT = 1
 DEFAULT_WKID = 4326
-DEFAULT_USER_AGENT = 'geopy/' + get_version()
+
+
+def get_default_user_agent():
+    return "geopy/" + gu.get_version()
+
 
 ERROR_CODE_MAP = {
     400: GeocoderQueryError,
@@ -73,7 +77,7 @@ class Geocoder(object): # pylint: disable=R0921
             scheme=DEFAULT_SCHEME,
             timeout=DEFAULT_TIMEOUT,
             proxies=None,
-            user_agent=DEFAULT_USER_AGENT
+            user_agent=None
         ):
         """
         Mostly-common geocoder validation, proxies, &c. Not all geocoders
@@ -87,7 +91,7 @@ class Geocoder(object): # pylint: disable=R0921
             )
         self.proxies = proxies
         self.timeout = timeout
-        self.headers = {'User-Agent': user_agent }
+        self.headers = {'User-Agent': user_agent or get_default_user_agent()}
 
         if self.proxies:
             install_opener(
@@ -172,12 +176,12 @@ class Geocoder(object): # pylint: disable=R0921
         else:
             status_code = None
         if status_code in ERROR_CODE_MAP:
-            raise ERROR_CODE_MAP[page.status_code]("\n%s" % decode_page(page))
+            raise ERROR_CODE_MAP[page.status_code]("\n%s" % gu.decode_page(page))
 
         if raw:
             return page
 
-        page = decode_page(page)
+        page = gu.decode_page(page)
 
         if deserializer is not None:
             try:
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -15,6 +15,7 @@ from geopy.compat import (
     ProxyHandler,
     URLError,
     install_opener,
+    Request,
 )
 from geopy.point import Point
 from geopy.exc import (
@@ -28,7 +29,7 @@ from geopy.exc import (
     GeocoderUnavailable,
     GeocoderParseError,
 )
-from geopy.util import decode_page
+from geopy.util import decode_page, get_version
 
 
 __all__ = (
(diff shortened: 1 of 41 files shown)


ijl:
I've cherry-picked 8b38ebb8918fa2b30bbe0b0ede6e1c3244c4946d for the moment, thanks.

geopy/util.py
```diff
@@ -80,3 +80,10 @@ def decode_page(page):
         else: # requests?
             encoding = page.headers.get("charset") or "utf-8"
             return str(page.content, encoding=encoding)
+
+
+def get_version():
```
ijl: `get_version()` should be replaced with use of a constant, to me.

geopy/util.py
```diff
@@ -80,3 +80,10 @@ def decode_page(page):
...
+    import geopy
+    return str(geopy.__version__)
```
ijl: `geopy.__init__.__version__` can move to `geopy.version.GEOPY_VERSION` or such to avoid circular imports, with it reimported in the package init for export, yeah?

geopy/geocoders/base.py
```diff
@@ -45,6 +46,11 @@
 DEFAULT_TIMEOUT = 1
 DEFAULT_WKID = 4326
 
+
+def get_default_user_agent():
+    return "geopy/" + gu.get_version()
```
ijl: This could be a constant.

ijl:
@sebastianneubauer it certainly makes sense, and thanks for working on it. Some comments. One significant thing is that every geocoder would need to accept and pass down a `user_agent` parameter.

sebastianneubauer:
As it it implemented now, if a geocoder has not implemented the acceptance and passing of a custom http header, the default value "geopy/VERSION" which is still much better than the urllib default value. As said, at the moment I implemented it only for the Nominatim geocoder, as it was requested by them. Of course it would make sense also for all or many other geocoders to implement this feature, but I fear,  I certainly would need help for this...Until then, I believe it improves also with this current implementation and maybe one could write an issue to implement this in other geocoders later on?

ijl:
Ok, I'll add it to other geocoders, then.

geopy/util.py
sebastianneubauer: You mean really an arbitrary constant like "1.0" or the version number of geopy? In my opinion for traceability and debugging it makes sense that it is the actual geopy package version? But in this case a constant would have to be held in sync manually...

ijl: Just that it can be `"geopy/%s" % geopy.version.GEOPY_VERSION` or the like, determined at interpreter start. No need for a function call.

geopy/util.py
sebastianneubauer: Good point, I will try that!

geopy/geocoders/base.py
sebastianneubauer: In my experience it is much nicer to test if dynamic things are not fixed at import time (constant), but at call time (function). It is then very easy to mock this function within the test to a known test return value, e.g: "mytestagent/1.2.3". At least I haven't figured out how to do that with a constant...
What would be the benefit of a constant?

ijl: Having it as a constant is cheaper. Not significant here, just the principle. You can mock anything, so, e.g., not tested: `with mock.patch('geopy.geocoders.base.DEFAULT_USER_AGENT', 'geopy/0.0.0')`. IIRC the usage below without a second parameter returns a MagickMock object, which is a callable instead of a string like you want.

geopy/util.py
sebastianneubauer: Hmm, wait, I cannot move geopy.__init__.__version__ as it it used by python distutils/setuptools to determine the version number for packaging. But maybe I find a better way to do it...

ijl: It's cool, you just need to put it in the init module to have it, e.g., `import geopy.version.GEOPY_VERSION as __version__`, yeah?

sebastianneubauer: Ha, nice thing, let's try!

sebastianneubauer: How is GEOPY_VERSION then set? It should be set by setup.py...

sebastianneubauer:
Ok, I think I managed to incorporate the suggestions, the default value of the user agent is now a constant and the circular import is removed...

sebastianneubauer:
Now added customizable user agents as optional parameter to all geocoders!
From my side this is finished for now...

ijl:
@sebastianneubauer ok thanks, I'll merge later.

geopy/geocoders/osm.py
```diff
@@ -156,8 +156,7 @@ def geocode(
             params = {'q': self.format_string % query}
 
         params.update({
-            # `viewbox` apparently replaces `view_box`
-            'viewbox': self.view_box,
+            'view_box': self.view_box,
```
ijl: Bad merge, I'll fix.

geopy/geocoders/placefinder.py
```diff
@@ -154,9 +154,6 @@ def geocode(
 
         :param valid_country_codes:
         :type valid_country_codes: list or tuple
-
-        :param bool with_timezone: Include the timezone in the response's
```
ijl: Bad merge, I'll fix.

ijl:
@sebastianneubauer sorry, this is stalled because it breaks Yahoo Placefinder, which uses requests. Would you be willing to fix that?

sebastianneubauer:
Oh, ok, but why are all tests passing? Would be much easier for me to have a breaking test for this bug....I will have a look....

sebastianneubauer:
@ijl : sorry, could you elaborate a bit further, what does "break" mean in detail. How can I reproduce this? Would it be an option to remove the user agent option from yahoo placefinder?

pnorman:
> My application that uses geopy was blocked by Nominatim as geopy uses the standard user agent header for requests provided by urllib (Python-urllib/2.7). Nominatim requires to add an identifying user agent header per application in order to be able to track and block requests if needed.

Just to note that the [Nominatim usage policy](http://wiki.openstreetmap.org/wiki/Nominatim_usage_policy) **requires** a valid HTTP Referer or User-Agent identifying the application and stock User-Agents as set by http libraries are explicitly not okay.

sebastianneubauer:
@ijl : what is the status here? this library violates the [Nominatim usage policy](http://wiki.openstreetmap.org/wiki/Nominatim_usage_policy) without this PR! So for us it is really urgent to have this included!

sebastianneubauer:
@ijl : Now I see, apparently I removed accidentally some code in a bad merge, I'll fix that...

sebastianneubauer:
ok, so I updated it and fixed the bad merges...still any open issues?

ijl:
@sebastianneubauer, Placefinder needed a fix in Base, but I added that. I'll release in 1.11.0 shortly. Thanks.

ijl:
@sebastianneubauer on PyPI.

sebastianneubauer:
wohooo, thanks!!

PR #121, opened 2015-04-18
Fix bug in OpenCage geocoder.

dthul:
The OpenCage geocoder ignores language and country settings at the moment and wrongfully only sets them when bounds is not None.
Seems to be a copy'n'paste bug.


ijl:
@dthul thanks but can you add regression tests?

ijl:
This was merged for 1.11.0.

PR #129, opened 2015-06-07
Classic haversine

saromanov:
Hi!
I just added calculation distance by classic haversine formula with method `great_circle_haversine` and in some cases it takes different results than `great_circle`. To be more clear, i runned this to methods on 6400 samples and in 53 percent of cases it takes different results(basically, a little different). Also, if to compare this with default calculation of distance (Vincenty), `great_circle_haversine` is more accurate than `great_circle`(In 70 percent of cases). 
I'm not documented it now, because i don't know, if you need it or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,13 @@ Using great-circle distance:
     >>> print(great_circle(newport_ri, cleveland_oh).miles)
     537.1485284062816
 
+Using great-circle haversine formula:
+    >>> from geopy.distance import great_circle_haversine
+    >>> newport_ri = (41.49008, -71.312796)
+    >>> cleveland_oh = (41.499498, -81.695391)
+    >>> print(great_circle_haversine(newport_ri, cleveland_oh).miles)
+    537.882275111
+
 ## Documentation
 
 More documentation and examples can be found at
--- a/geopy/distance.py
+++ b/geopy/distance.py
@@ -281,6 +281,19 @@ class great_circle(Distance):
 
         return Point(units.degrees(radians=lat2), units.degrees(radians=lng2))
 
+class great_circle_haversine(great_circle):
+    def __init__(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


ijl:
@saromanov, I'm ok including this. Could you document it more, and add more specific regression tests?

saromanov:
Yes, i will do it in the next few days.

saromanov:
Document great circle haversine formula and provided additional tests

cffk:
Something fishy is going on here...

When I run the example in README.md I get precisely the same results
from great_circle and great_circle_haversine, i.e.,537.1485284062816
miles.  And yet the README.md claims a *different* result for
great_circle_haversine, namely 537.882275111 miles (which happens to be
wrong!).

Also, why the comarison with Vincenty?  One computes the distance on the
sphere and the other on the ellipsoid.

The current formula used by great_circle is just as accurate as the
haversine formula for distances up to a quarter meridian and *more*
accurate than the haversine formula for distances between a quarter and a
half meridian.  For example:

great_circle reports the distance between (-30,0) and (29.999,179.999)
as 20020.578816010657 km.

great_circle_haversine reports the distance as 20020.57881599349 km.

The actual distance (for radius 6372.795 km) is 20020.578816010657667 km
+/- 1 picometer.  So the error in the great_circle result is about 1
nanometer and the errror in the great_circle_haversine result is about
17 micrometers.

In addition, there's no point in junking up the user interface with two
equivalent methods of calculating the same quantity.

So unless there's some fundamental gap in my understanding (which is
entirely possible), I recommend sticking with the existing great_circle
method.

KostyaEsmukov:
Also linking #144 which should provide a better accuracy for `distance`.

KostyaEsmukov:
It seems to be abandoned, so I'm closing this. But feel free to chime in if it's still relevant.

PR #131, opened 2015-06-21
Update googlev3.py

wimfeijen:


ijl:
Fixed in another commit to avoid a generic commit message. Thanks.

PR #134, opened 2015-07-09
Adding support for extra parameters

steinnes:
This way we can use third party services like PickPoint like so:

```
>>> from geopy import geocoders
>>> g = geocoders.Nominatim(domain='pickpoint.io/api/v1',
extra_params={'key': '....'})
>>> g.reverse("64.0, 0.0")
Location((0.0, 0.0, 0.0))
```

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -192,7 +192,6 @@ class Nominatim(Geocoder):
 
         url = "?".join((self.api, urlencode(params)))
         logger.debug("%s.geocode: %s", self.__class__.__name__, url)
-        print self._call_geocoder
         return self._parse_json(
             self._call_geocoder(url, timeout=timeout), exactly_one
         )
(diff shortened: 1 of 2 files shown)


steinnes:
Are there any objections?  I am perfectly willing to change my approach so that it's more in-line with the rest of the project structure (which I am not familiar with yet)? :-)

ijl:
I've just added a section on additional parameters to https://github.com/geopy/geopy/blob/master/CONTRIBUTING.md. Can you make it match that? I'll close this in case you've given up after the delay. Feel free to reopen.

steinnes:
Ah, I can change it, but of course I am just pip installing my version right now.  The reason I decided to add this in the constructor was because I thought it would be tedious to litter my code with the authentication key, and I assumed there were other services out there which had a similar key system :-)

If not, and this is a total corner case, I'll be happy to adapt my PR.  As long as my work saves someone else some time sometime ;-)

ijl:
@steinnes key would be useful, and is done on the init in other geocoders. If it was a specific named arg it would be good.

PR #142, opened 2015-08-11
FIX: index out of range if no results found (yandex)

facciocose:


ijl:
Merged and to be released in 1.11.0, thanks.

PR #143, opened 2015-08-12
Remove duplicate assignment

isedwards:
Line duplicated twice (see three lines above)
`self.country_bias = country_bias`


ijl:
Merged and to be released in 1.11.0, thanks.

PR #150, opened 2015-09-01
Use the exact conversion factor from kilometers to miles.

cffk:
This should be uncontroversial.  Possibly a few tests will fail.
But this can be easily fixed.


ijl:
Merged in 1963a90f6358c12a6beb58d92b366021f3abf9c9, thanks.

PR #151, opened 2015-09-01
Use the IUGG mean earth radius for EARTH_RADIUS.

cffk:
The previous value was the "quadratic mean radius" which is defined here

  http://math.wikia.com/wiki/Ellipsoidal_quadratic_mean_radius

and was, for a time, touted on Wikipedia.  Here there's a flaw in the
derivation of this quantity -- the way the azimuths was sampled was
wrong; see

  https://en.wikipedia.org/wiki/Talk:Great-circle_distance/Archive_1

The IUGG value minimizes the RMS difference between the spherical and
ellipsoidal distances.

### DIFF
--- a/geopy/distance.py
+++ b/geopy/distance.py
@@ -8,10 +8,12 @@ formulas, with a default of Vincenty available as the function
 `geopy.distance.distance`.
 
 Great-circle distance (:class:`.great_circle`) uses a spherical model of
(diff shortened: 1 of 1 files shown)


ijl:
Merged in 9a44bc33aaf6966a2a22516f0ba7fd0b1035d84e, thanks.

PR #153, opened 2015-09-14
Fix the name parsing in Photon geocoder

mthh:
To correct a mistake when retrieving the location name on Photon geocoder.
It will now output as expected:

``` python
In [15]: photon.geocode('Duingen', osm_tag='place')
Out[15]: Location(Duingen, Lower Saxony, Germany, (52.0057957, 9.6970268, 0.0))
```


KostyaEsmukov:
This has been fixed in 2bb4a92, so I'm going to close this. But thank you anyway!

PR #154, opened 2015-09-18
Add support for Teleport Geocoder

magnushiie:
I implemented support for forward and reverse geocoding using [Teleport APIs](https://developers.teleport.org/api/).

Teleport geocoding APIs are based on GeoNames cities1000 (i.e. cities with population > 1000) dataset, but with different logic for searching for cities by name, administrative division and country.

It also integrates with the Teleport-specific data - e.g. it's possible to use the `forward_embeddings` and `reverse_embeddings` parameters to specify that the returned city information should also be accompanied by the Teleport City scores using the `"city:item/city:urban_area/ua:scores"` embedded relation as shown in the `Teleport` class documentation.

Disclaimer: I work for @teleport

Let me know if there's anything that needs to be changed in order to merge.

### DIFF
--- a/geopy/geocoders/teleport.py
+++ b/geopy/geocoders/teleport.py
@@ -11,7 +11,10 @@ from geopy.geocoders.base import (
 )
 from geopy.location import Location
 from geopy.util import logger
-from urllib import quote
+try:
+    from urllib.parse import quote  # @UnresolvedImport @UnusedImport
+except:
+    from urllib import quote  # @Reimport
 
 
 __all__ = ("Teleport", )
@@ -24,6 +27,12 @@ DEFAULT_REVERSE_EMBEDDINGS = (
     "location:nearest-cities/location:nearest-city" +
     "/{city:country,city:admin1_division}"
 )
+
+
+try:
+    basestring
+except:
+    basestring = str  # @ReservedAssignment
 
 
 class Teleport(Geocoder):
--- a/geopy/geocoders/teleport.py
+++ b/geopy/geocoders/teleport.py
@@ -1,0 +1,274 @@
+"""
+:class:`Teleport` geocoder.
+"""
+
+from geopy.compat import urlencode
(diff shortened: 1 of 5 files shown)


magnushiie:
The failures seem to be caused by OpenMapQuest no longer accepting requests without API Key.

magnushiie:
I can remove the merge of the OpenMapQuest change from this pull request if needed.

test/geocoders/teleport.py
```diff
@@ -0,0 +1,42 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.delta = 0.04
+
+    def test_unicode_name(self):
+        """
+        Teleport.geocode unicode
+        """
+        # work around ConfigurationError raised in GeoNames init
+        self.geocoder = Teleport(scheme='http')
```
KostyaEsmukov: I think of switching all tests to https. Almost all of them (if not exactly all) support https these days. So if this is the case for Teleport, could you please remove the `scheme` kwarg, to use the default (`https`) scheme?

geopy/geocoders/teleport.py
```diff
@@ -0,0 +1,283 @@
...
+
+    See API documentation at (https://developers.teleport.org/api/) for
+    details.
+
+    Example::
+
+        from geopy.geocoders import Teleport
+        from geopy.geocoders.teleport import DEFAULT_FORWARD_EMBEDDINGS
+
+        teleport = Teleport(forward_embeddings=DEFAULT_FORWARD_EMBEDDINGS +
```
KostyaEsmukov: Could you please add this example to tests? I mean the `forward_embeddings` and `Teleport.get_embedded` parts, which are not covered in tests yet.

magnushiie:
I don't work for Teleport any more. I'll let @joonathan decide whether to continue with this.

magnushiie:
Closing for now, @teleport can re-create a new PR.

PR #155, opened 2015-09-18
adds api key to mapquest request

pimlinders:
Since September 9th it appears map quest now requires an API key for nominatim requests. In openmapquest.py it states "No API Key is needed by the Nominatim based - platform." but this is no longer true. Removed the comment. Added the api key to the params.


KostyaEsmukov:
Fixed by 25c5c48. There were 4 PRs addressing the same issue: #155, #160, #169, #228, but neither was ready enough to be merged straight away. I'm going to close this, but thank you!

PR #156, opened 2015-09-18
Make OpenMapQuest tests conditional on environment variable

magnushiie:
Tests seem to recently fail due to OpenMapQuest no longer accepting key-less API requests.

### DIFF
--- a/test/geocoders/openmapquest.py
+++ b/test/geocoders/openmapquest.py
@@ -5,6 +5,15 @@ from test.geocoders.util import GeocoderTestBase, env
 import unittest
 
 
+class OpenMapQuestNoNetTestCase(GeocoderTestBase):  # pylint: disable=R0904,C0111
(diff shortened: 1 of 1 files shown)


riordan:
Having done a bit of digging, it appears this pull request will fix the issue with the build not playing nice with TravisCI (#176). That should allow other PR's to start coming in.

It should be noted that a minor change in documentation should follow merging this, noting that Mapquest Open does now require an API key (this has been done in several PR's since, notably #155 #160 #169).

rodenj:
We (@kingb and I) had an interest in using an API key with OpenMapQuest (issue #168), but to get the existing tests to work we also needed to add the OPENMAPQUEST_APIKEY variable to the key list for the environment within Util to find it.  Prior to pull request #228, the api key could not be retrieved with env.get.

PR #166, opened 2015-11-28
Additional parameters for Bing reverse method

brot:
Optional parameters that influence the form of
the response such as culture (sets the language
of the response) or whether to include the
neighborhood or the two-letter ISO code of the
country.

### DIFF
--- a/geopy/geocoders/bing.py
+++ b/geopy/geocoders/bing.py
@@ -164,7 +164,14 @@ class Bing(Geocoder):
             exactly_one
         )
 
-    def reverse(self, query, exactly_one=True, timeout=None):
+    def reverse(
+            self,
+            query,
+            exactly_one=True,
+            timeout=None,
+            culture=None,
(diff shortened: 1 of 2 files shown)


brot:
Would be nice if someone could check this pull-request. I would like to see this change in the upstream code.

brot:
Is there any change to get this pull request accepted? Is this project still active? At the moment there are 27 pull request open. This is one of them and it's open since Nov 2015!

brot:
I added some tests for the two new parameters. But it seems that there are no API keys configured in the CI environment. On my local machine with my Bing API key the tests run fine

KostyaEsmukov:
The reason why API key is missing in CI is because Travis doesn't expose environment variables to forks due to security considerations. I haven't found a decent solution for this yet, unfortunately.

But I ran the tests locally and I confirm that they have passed.

LGTM, merging. Thank you!

PR #172, opened 2016-01-28
Added support for ArcGIS authentication with OAuth client ID and clie…

nh-cadmus:
I had trouble getting the ArcGIS service to work when authenticating with username/password, and would prefer not to keep those credentials in a program, so I added support to the ArcGIS service for using OAuth2 with client ID and secret.  

Added client_id and client_secret arguments to ArcGIS constructor, allowing users to authenticate using OAuth instead of entering their username and password. Added checks to ensure that username/password auth isn't attempted at the same time as OAuth.

### DIFF
--- a/geopy/geocoders/arcgis.py
+++ b/geopy/geocoders/arcgis.py
@@ -26,11 +26,12 @@ class ArcGIS(Geocoder):  # pylint: disable=R0921,R0902,W0223
     _TOKEN_EXPIRED = 498
     _MAX_RETRIES = 3
     auth_api = 'https://www.arcgis.com/sharing/generateToken'
+    oauth_api = 'https://www.arcgis.com/sharing/oauth2/token'
 
     def __init__(self, username=None, password=None, referer=None, # pylint: disable=R0913
                  token_lifetime=60, scheme=DEFAULT_SCHEME,
                  timeout=DEFAULT_TIMEOUT, proxies=None,
-                user_agent=None):
+                user_agent=None, client_id=None, client_secret=None):
         """
         Create a ArcGIS-based geocoder.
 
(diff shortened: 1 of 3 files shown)


nh-cadmus:
It looks like the build failures were not introduced by my changes - in fact, the _master_ branch is failing right now too.

geopy/geocoders/arcgis.py
```diff
@@ -62,26 +63,52 @@ def __init__(self, username=None, password=None, referer=None, # pylint: disable
...
         super(ArcGIS, self).__init__(
             scheme=scheme, timeout=timeout, proxies=proxies, user_agent=user_agent
         )
-        if username or password or referer:
-            if not (username and password and referer):
-                raise ConfigurationError(
-                    "Authenticated mode requires username,"
-                    " password, and referer"
-                )
+        if username or password or referer or client_id or client_secret:
```
KostyaEsmukov: I think it would be cleaner to introduce these two vars:

```python
is_oauth = client_id or client_secret
is_username_auth = username or password or referer
```

geopy/geocoders/arcgis.py
```diff
@@ -218,19 +245,36 @@ def _refresh_authentication_token(self):
...
+                'f': 'json'
+            }
+        elif self.client_id:
+            auth_url = self.oauth_api
+            token_request_arguments = {
+                'client_id': self.client_id,
+                'client_secret': self.client_secret,
+                # User/pass method takes token lifetime in seconds, 
+                # OAuth method takes token lifetime in minutes
+                'expiration': self.token_lifetime / 60,
```
KostyaEsmukov: Shouldn't it be rounded to int? In Python 3 division result is `float`, but in Python 2 it's `int`.

test/geocoders/arcgis.py
```diff
@@ -98,3 +105,29 @@ def test_basic_address(self):
             {"query": "Potsdamer Platz, Berlin, Deutschland"},
             {"latitude": 52.5094982, "longitude": 13.3765983},
         )
+        
+        
+@unittest.skipUnless(  # pylint: disable=R0904,C0111
+    env.get('ARCGIS_USERNAME') is not None \
```
KostyaEsmukov: I believe this condition should refer to `ARCGIS_CLIENT_ID` and `ARCGIS_CLIENT_SECRET` instead of `ARCGIS_USERNAME`/etc.

KostyaEsmukov:
Closing due to lack of feedback.

PR #174, opened 2016-02-10
Add reverse geocoding support for OpenMapQuest

janpascal:
Add reverse geocoding support for openmapquest. Added some more robustness to its json parser to fix crashes due to unexpected messages from openmapquest.

### DIFF
--- a/geopy/geocoders/openmapquest.py
+++ b/geopy/geocoders/openmapquest.py
@@ -163,10 +163,10 @@ class OpenMapQuest(Geocoder): # pylint: disable=W0223
         """
         Return location and coordinates tuple from dict.
         """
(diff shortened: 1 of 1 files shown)


KostyaEsmukov:
Thanks! I'm holding this, because I'd prefer instead of improving the current OpenMapQuest implementation to make it extend the `Nominatim` class (similarly to `Pickpoint`). However OpenMapQuest [runs on a somewhat older version](https://developer.mapquest.com/forum/differences-between-openmaquestcom-and-nominatimopenstreetmaporg) of Nominatim, so some of the `Nominatim` class parameters might not work for OpenMapQuest.

Let me know if this is a severe issue for anyone.

PR #177, opened 2016-02-23
Feature to use Nominatim with 3rd party services (e.g. PickPoint)

romaintha:
Add 3 parameters to geocoder Nominatim  in order to be used with third party services such as PickPoint


KostyaEsmukov:
Thank you for this. 

I'm closing it for the following reasons:
1. PickPoint has been added as a separate geocoder class in #246.
2. IMO as original Nominatim doesn't use API keys, its class shouldn't use them too.
3. It's more practical to introduce new geocoders by subclassing rather than creating an instance with custom domain and forward/reverse URL parts, because a class is more reusable and customizable.

PR #179, opened 2016-02-25
Coverage 4.0 doesn't support Python 3.2

### LINKED ISSUE #178: Python 3.2 coverage has a syntax error
which again causes Travis CI tests to fail (#176)
https://travis-ci.org/geopy/geopy/jobs/111224816

The root cause seems to be that the latest version of coverage (4.1b2) has a syntax error
https://github.com/nedbat/coveragepy/blob/master/coverage/annotate.py#L85 

introduced by adding utf8 support here
https://github.com/nedbat/coveragepy/commit/fdfdb2432dcf9e544e7c6a70421e74e5c56adf45

The standard action seems to be to use coverage 3.7.1 on Python 3.2 or drop Python 3.2 testing (see https://github.com/menegazzo/travispy/issues/20)

jqnatividad:
On Python 3.2, use earlier version of coverage (v3.7.1)

Fixes #178.

see https://bitbucket.org/ned/coveragepy/issues/407/coverage-failing-on-python-325-using

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ python:
     - "3.4"
     - "pypy"
     - "pypy3"
+install:
+    # Coveralls 4.0 doesn't support Python 3.2
+    - if [ "$TRAVIS_PYTHON_VERSION" == "3.2" ]; then travis_retry pip install coverage==3.7.1; fi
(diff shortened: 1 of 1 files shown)


riordan:
You pulled the sword from the Travis-CI. YOU ARE THE CHOSEN ONE!

icemac:
+1 to merge this PR to get rid of the failing Travis-CI.

PR #182, opened 2016-03-09
Add restricting results via bounded parameter in Nominatim

0x836:
Viewbox area is **preferred** area to find search results.
The goal of adding bounded parameter is guaranteed restricting results to the bounding box.

For example
[query without bounded](http://nominatim.openstreetmap.org/search.php?q=строитель томск&viewbox=84.719353,56.588456,85.296822,56.437293)
geolocator = Nominatim(format_string='%s,Томск', view_box=('84.719353','56.588456','85.296822','56.437293')) - I want to get coords from Tomsk/Tomsk district etc excluding cities/villages south of the Tomsk

so [query with bounded](http://nominatim.openstreetmap.org/search.php?q=строитель томск&viewbox=84.719353,56.588456,85.296822,56.437293&bounded=1) is what I need

geolocator = Nominatim(format_string='%s,Томск', view_box=('84.719353','56.588456','85.296822','56.437293'), bounded=True)

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -38,6 +38,7 @@ class Nominatim(Geocoder):
             self,
             format_string=DEFAULT_FORMAT_STRING,
             view_box=None,
+            bounded=False,
             country_bias=None,
             timeout=DEFAULT_TIMEOUT,
             proxies=None,
@@ -52,6 +53,9 @@ class Nominatim(Geocoder):
             is just '%s'.
 
         :param tuple view_box: Coordinates to restrict search within.
+
+        :param bool bounded: Restrict the results to only items contained
+            with the bounding view_box.
 
         :param string country_bias: Bias results to this country.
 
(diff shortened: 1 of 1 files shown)


senjed:
I added the bounded parameter in a similar way and I get None as a result of geocoding. any updates on this issue?

KostyaEsmukov:
Merged in #308. Thanks @0x836 !

@senjed The `None` you're getting might be because Nominatim is unable to satisfy your query within the view_box.

senjed:
well I don't think that the None that I am getting is because of that since the view_box that I use is for a city and when I include the name of the city in the query (but without view_box)  I will get relevant results with coordinates inside the view_box.

KostyaEsmukov:
That is interesting. But anyway, this looks like a peculiarity in Nominatim's behavior, which doesn't seem to be related to geopy itself (and this PR in particular). I would say that your #307 issue is a better venue for that (if it's still a problem).

PR #183, opened 2016-03-13
Added support for Mapzen geocoder service

migurski:
Includes tests; Travis failures match those on master.

* [x] Get tests which pass locally to work on Travis.

### DIFF
--- a/geopy/geocoders/__init__.py
+++ b/geopy/geocoders/__init__.py
@@ -91,6 +91,7 @@ __all__ = (
     'Yandex',
     "What3Words",
     "Photon",
+    "Mapzen",
 )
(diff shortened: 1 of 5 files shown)


eldang:
Thank you for making this!  FWIW, I've been able to use it to connect to the Mapzen geocoder by just saving mapzen.py locally and importing Mapzen from it, with good results.

migurski:
Glad it works for you, Eldan!

KostyaEsmukov:
Mapzen has shut down their API, so I'm thinking about removing this code eventually. If you think that it should stay – please join the discussion at #258. Thanks.

PR #184, opened 2016-04-13
Fixed headers being not set

KostyaEsmukov:
Passing custom `user_agent` doesn't work. This fixes it. I think it was a simple typo.

Not sure how to properly test it though. Maybe mocking `self.urlopen`?


KostyaEsmukov:
Apparently the fix in this PR breaks some geocoders which pass `Request` instead of a string. The proper fix is at 3395aee.

PS 2 years later I'm rejecting my own PR. This has never happened to me before. :)

PR #194, opened 2016-06-07
Fix error in Vincenty destination function, closes #186.

### LINKED ISSUE #186: Error in vincenty destination function
I've reviewed the code of the vincenty distance and destination function against the mathematical formulas cited on the Wikipedia page. From what I can tell there is one minor error in the destination function, which fails to do a power2 on one term.

Currently delta_sigma inside the loop (https://github.com/geopy/geopy/blob/master/geopy/distance.py#L482) is defined as:

```
delta_sigma = B * sin_sigma * (
    cos2_sigma_m + B / 4. * (
        cos_sigma * (
            -1 + 2 * cos2_sigma_m
        ) - B / 6. * cos2_sigma_m * (
            -3 + 4 * sin_sigma ** 2
        ) * (
            -3 + 4 * cos2_sigma_m ** 2
        )
    )
)
```

I believe the `-1 + 2 * cos2_sigma_m` part is missing a power 2 and should be `-1 + 2 * cos2_sigma_m ** 2` instead, so it all reads:

```
delta_sigma = B * sin_sigma * (
    cos2_sigma_m + B / 4. * (
        cos_sigma * (
            -1 + 2 * cos2_sigma_m ** 2
        ) - B / 6. * cos2_sigma_m * (
            -3 + 4 * sin_sigma ** 2
        ) * (
            -3 + 4 * cos2_sigma_m ** 2
        )
    )
)
```

hannosch:
The test uses the same example as given by movable-type.co.uk and mozilla/ichnaea and matches the output of geographiclib's geosolve online calculator. The test fails without the fix.

@jqnatividad any comments?

### DIFF
--- a/geopy/distance.py
+++ b/geopy/distance.py
@@ -482,7 +482,7 @@ class vincenty(Distance):
             delta_sigma = B * sin_sigma * (
                 cos2_sigma_m + B / 4. * (
                     cos_sigma * (
-                        -1 + 2 * cos2_sigma_m
+                        -1 + 2 * cos2_sigma_m ** 2
                     ) - B / 6. * cos2_sigma_m * (
                         -3 + 4 * sin_sigma ** 2
                     ) * (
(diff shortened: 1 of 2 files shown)


cffk:
I recommend merging this pull request.  It fixes an outright bug.

PR #204, opened 2016-07-29
Fix tests

medecau:
This pull-request is an effort to get all tests passing. On both tox and travis-ci.

NaviData is gone as it's API was returning HTTP status code 404. If the API comes back online the old code can be recovered from history.
*I mistyped the name as "navigate" on one of the commit messages.*

tox config command for nosetests was using the parameter `--processes=-1`. I have not investigated the cause but it was resulting in some failed tests. I removed it. One can still achieve parallelism in tests using the `detox` tool.

I added pypy and pypy3 to the tox config. Use `--skip-missing-interpreters` if you don't have these installed.

I also added python 3.5 to both tox and travis config files.


jqnatividad:
Ugh!  Now What3Words tests are failing :( 

It seems What3Words just updated their API...
https://docs.what3words.com/api/v2/

PR #216, opened 2016-09-09
fix geopy tests

aqueiros:
fixing failing GoogleV3 test


jqnatividad:
This is great stuff! However, Python 3.2 tests are still failing...

PR #226, opened 2016-10-15
Convert What3Words Geocoder to the v2 API.

jmb:
Also remove the OneWord functionality that has been deprecated by
What3Words.

### DIFF
--- a/geopy/geocoders/what3words.py
+++ b/geopy/geocoders/what3words.py
@@ -38,7 +38,7 @@ class What3Words(Geocoder):
             user_agent=None,
     ):
         """
-        Initialize a What3Words geocoder with 3-word or OneWord-address and
+        Initialize a What3Words geocoder with 3-word address and
         What3Words API key.
 
             .. versionadded:: 1.5.0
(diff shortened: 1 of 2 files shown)


tym-oao:
Note: what3words API v1 will no longer be available after November 30. It would be good to get this merged before then.

geopy/geocoders/what3words.py
```diff
@@ -53,7 +52,9 @@ def __init__(
 
         :param string scheme: Use 'https' or 'http' as the API URL's scheme.
             Default is https. Note that SSL connections' certificates are not
-            verified.
+            verified.  As of API v2 only 'https' is supported - this parameter
+            remains for backwards compatibility (setting to 'http' will stop
+            this geocoder working).
```
KostyaEsmukov: I suggest to explicitly raise an exception in case `http` is passed.

```
if scheme == "http":
    raise ConfigurationError("What3Words doesn't support `http` scheme anymore. Please use `https` instead.")
```

geopy/geocoders/what3words.py
```diff
@@ -139,57 +138,27 @@ def _parse_json(self, resources, exactly_one=True):
         Parse type, words, latitude, and longitude and language from a
         JSON response.
         """
-        if resources.get('error') == "X1":
-            raise exc.GeocoderAuthenticationFailure()
 
-        if resources.get('error') == "11":
+        if 'code' in resources['status']:
             raise exc.GeocoderQueryError(
```
KostyaEsmukov: Could you please raise `GeocoderAuthenticationFailure` here when `resources['status']['code'] == 401`?

And the same for `reverse` method below.

https://docs.what3words.com/api/v2/#errors

KostyaEsmukov:
Merged in #303

KostyaEsmukov:
Thanks for this!

tym-oao:
Yowsah!

PR #227, opened 2016-10-17
Update ArcGIS to latest API

jmb:
As per
https://developers.arcgis.com/rest/geocode/api-reference/geocoding-find-
address-candidates.htm the `find` operation has been deprecated. This
migrates to the `findAddressCandidates` operation but utilising the
`singleLine` functionality which mirrors the old `find` operation.

### DIFF
--- a/geopy/geocoders/arcgis.py
+++ b/geopy/geocoders/arcgis.py
@@ -109,7 +109,7 @@ class ArcGIS(Geocoder):  # pylint: disable=R0921,R0902,W0223
         )
         return self._base_call_geocoder(request, timeout=timeout)
 
-    def geocode(self, query, exactly_one=True, timeout=None):
+    def geocode(self, query, exactly_one=True, out_fields=None, timeout=None):
         """
         Geocode a location query.
 
@@ -117,6 +117,9 @@ class ArcGIS(Geocoder):  # pylint: disable=R0921,R0902,W0223
 
         :param bool exactly_one: Return one result or a list of results, if
             available.
+
+        :param string out_fields: Comma-separated list of output fields to be returned in the
+            attributes field of the raw data.
 
         :param int timeout: Time, in seconds, to wait for the geocoding service
             to respond before raising a :class:`geopy.exc.GeocoderTimedOut`
@@ -126,6 +129,8 @@ class ArcGIS(Geocoder):  # pylint: disable=R0921,R0902,W0223
         params = {'singleLine': query, 'f': 'json'}
         if exactly_one is True:
             params['maxLocations'] = 1
+        if out_fields is not None:
+            params['outFields'] = out_fields
         url = "?".join((self.api, urlencode(params)))
         logger.debug("%s.geocode: %s", self.__class__.__name__, url)
         response = self._call_geocoder(url, timeout=timeout)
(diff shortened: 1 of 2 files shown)


geopy/geocoders/arcgis.py
```diff
@@ -118,14 +118,27 @@ def geocode(self, query, exactly_one=True, timeout=None):
...
         """
-        params = {'text': query, 'f': 'json'}
+        params = {'singleLine': query, 'f': 'json'}
         if exactly_one is True:
             params['maxLocations'] = 1
+        if out_fields is not None:
+            # Python 2/3 compatibility
+            if 'basestring' not in globals():
+                basestring = str
+            if isinstance(out_fields, basestring):
```
diogocp: You can use `geopy.compat.string_compare` here: `if isinstance(out_fields, string_compare):`

test/geocoders/arcgis.py
```diff
@@ -63,6 +63,34 @@ def test_unicode_name(self):
...
+    def test_geocode_with_out_fields_string(self):
+        """
+        ArcGIS.geocode with outFields string
+        """
+        self.geocode_run(
+            {"query": "Heathrow Airport", "out_fields":"Country"},
+            {"raw": {
+                u'attributes': {u'Country': u'GBR'},
+                u'score': 100, u'location': {u'y': 51.47114986100047, u'x': -0.45648655399958216},
+                u'extent': {u'xmin': -0.481488, u'ymin': 51.44615, u'ymax': 51.49615, u'xmax': -0.431488},
```
diogocp: It might be a bad idea to test the exact coordinates/extent of the airport, since those might change if the expansion plans go ahead :)

geopy/geocoders/arcgis.py
```diff
@@ -118,14 +118,27 @@ def geocode(self, query, exactly_one=True, timeout=None):
         :param bool exactly_one: Return one result or a list of results, if
             available.
 
+        :param out_fields:
+        :type out_fields: A list of output fields to be returned in the
+            attributes field of the raw data. This can be either a python
+            list/tuple of fields or a comma-separated string.
+
```
diogocp: A link to the ArcGIS documentation would be useful here:
> See https://developers.arcgis.com/rest/geocode/api-reference/geocoding-service-output.htm for a list of supported output fields. If you want to return all supported output fields, set out_fields="*".

jmb: Good idea, thanks!

test/geocoders/arcgis.py
jmb: Hehe, I'll choose somewhere else! ;)

geopy/geocoders/arcgis.py
```diff
@@ -109,7 +109,7 @@ def _authenticated_call_geocoder(self, url, timeout=None):
         )
         return self._base_call_geocoder(request, timeout=timeout)
 
-    def geocode(self, query, exactly_one=True, timeout=None):
+    def geocode(self, query, exactly_one=True, out_fields=None, timeout=None):
```
KostyaEsmukov: Could you please put the `out_fields` argument after `timeout`? Otherwise it would break current usages of `geocode` call, where timeout is passed as a positional arg. Also the base method has `timeout` too, so we should stick to that order.

KostyaEsmukov:
I've created another PR based on this, because git push to this branch is not allowed to me.

@jmb thank you for this! Closing in favor of #287

Issue #168, opened 2015-12-22
OpenMapQuest needs api_key


dpwilt:
Open MapQuest now requires an API key, which if supplied is not used by the geocode routine.  I think the fix is just to add:

```
    params['key'] = self.api_key 
```

before the line:

```
    url = "&".join((self.api, urlencode(params)))
```

in the geocode() class method within the openmapquest.py file.

Dan

kingb:
We finally figured out the problem... @rodenj will be submitting a pull request that should pass tests shortly.

KostyaEsmukov:
Fixed by 25c5c48

PR #169, opened 2016-01-12
Add api_key in OpenMapQuest geocoder

dcere:
Should fix issue #168

PR #160, opened 2015-10-15
add api key to param for OpenMapQuest Geocoder

j-san:
api key is now a mandatory param (http://open.mapquestapi.com/geocoding/) so just send it to the api request :tada:. 
Doesn't works any more without this param since MapQuest licensing structure change (http://devblog.mapquest.com/2014/11/17/are-you-a-community-edition-licensed-data-user/).

PR #228, opened 2016-10-21
Fix for OpenMapQuest issue geopy/geopy#168

rodenj:
Added missing OPENMAPQUEST_APIKEY to util.py key list so it can be read from the environment; And added use of that api key within OpenMapQuest base api URL.


KostyaEsmukov:
Fixed by 25c5c48. There were 4 PRs addressing the same issue: #155, #160, #169, #228, but neither was ready enough to be merged straight away. I'm going to close this, but thank you!

Issue #239, opened 2017-04-11
GeocodeFarm throws an exception when there are no results


pembo13:
# Sample query
`342002`

# Sample returned status:

```javascript
"STATUS": {
	"access": "KEY_VALID, ACCESS_GRANTED",
	"status": "FAILED, NO_RESULTS"
}
```

# Expected Result

`None`

# Actual Result

`GeocoderServiceError: KEY_VALID, ACCESS_GRANTED`

PR #240, opened 2017-04-11
Add check for a NO_RESULT to GeocodeFarm

pembo13:
Issue  #239

Added two lines to check for the `NO_RESULTS` status from webservice and return `None` instead of an exception.

### DIFF
--- a/geopy/geocoders/geocodefarm.py
+++ b/geopy/geocoders/geocodefarm.py
@@ -146,8 +146,6 @@ class GeocodeFarm(Geocoder):
         geocoding_results = api_result["geocoding_results"]
         self._check_for_api_errors(geocoding_results)
 
-        if "NO_RESULTS" in geocoding_results.get("STATUS", {}).get("status", ""): return None
-
         places = self.parse_code(geocoding_results)
         if exactly_one is True:
             return places[0]
@@ -161,7 +159,6 @@ class GeocodeFarm(Geocoder):
         in the api response.
         """
         status_result = geocoding_results.get("STATUS", {})
-        if "NO_RESULTS" in status_result.get("status", ""): return
         api_call_success = status_result.get("status", "") == "SUCCESS"
         if not api_call_success:
             access_error = status_result.get("access")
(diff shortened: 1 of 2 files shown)


pembo13:
@KostyaEsmukov test added in 6d8d023b4a847cdf6c0352d009e4141b28da7be9

geopy/geocoders/geocodefarm.py
```diff
@@ -159,6 +161,7 @@ def _check_for_api_errors(geocoding_results):
         in the api response.
         """
         status_result = geocoding_results.get("STATUS", {})
+        if "NO_RESULTS" in status_result.get("status", ""): return
```
KostyaEsmukov: Is there any reason to have this check here? If there's none, I suggest to remove this line.

As far as I can see the following two lines shouldn't misbehave even when the status contains `NO_RESULTS`.

test/geocoders/geocodefarm.py
```diff
@@ -96,6 +96,35 @@ def mock_call_geocoder(*args, **kwargs):
         with self.assertRaises(exc.GeocoderQuotaExceeded):
             self.geocoder.geocode('435 north michigan ave, chicago il 60611')
 
+    def test_no_results(self):
+        """
+        GeocodeFarm unhandled error
```
KostyaEsmukov: This is clearly a copy-paste from another test. You can remove this comment completely (instead of updating it), because after switch to pytest (#266) they all will be removed anyway.

geopy/geocoders/geocodefarm.py
pembo13: Sorry, if I'm being honest this was long enough ago that I forgot my rational for the double lines...  I'll figure that out and cleanup.

test/geocoders/geocodefarm.py
pembo13: Yes, I wanted to emulate existing tests instead of reinventing unnecessarily... I was sloppy and missed that comment. Will remove.

geopy/geocoders/geocodefarm.py
pembo13: @KostyaEsmukov I double checked, both checks are necessary.

test/geocoders/geocodefarm.py
pembo13: @KostyaEsmukov removed that comment

PR #246, opened 2017-06-26
Add PickPoint geocoder.

cylon-v:
This PR adds PickPoint geocoder, which is commercial version of Nominatim hosted at https://pickpoint.io.

### DIFF
--- a/geopy/geocoders/__init__.py
+++ b/geopy/geocoders/__init__.py
@@ -90,6 +90,7 @@ __all__ = (
     "Mapzen",
     "OpenCage",
     "OpenMapQuest",
+    "PickPoint",
     "Nominatim",
     "Photon",
     "YahooPlaceFinder",
--- a/geopy/geocoders/__init__.py
+++ b/geopy/geocoders/__init__.py
@@ -113,6 +113,7 @@ from geopy.geocoders.opencage import OpenCage
 from geopy.geocoders.openmapquest import OpenMapQuest
 from geopy.geocoders.osm import Nominatim
 from geopy.geocoders.photon import Photon
+from geopy.geocoders.pickpoint import PickPoint
 from geopy.geocoders.placefinder import YahooPlaceFinder
 from geopy.geocoders.smartystreets import LiveAddress
 from geopy.geocoders.what3words import What3Words
@@ -136,6 +137,7 @@ SERVICE_TO_GEOCODER = {
     "mapzen": Mapzen,
     "opencage": OpenCage,
     "openmapquest": OpenMapQuest,
+    "pickpoint": PickPoint,
     "nominatim": Nominatim,
     "photon": Photon,
     "yahoo": YahooPlaceFinder,
@@ -165,4 +167,3 @@ def get_geocoder_for_service(service):
             "Unknown geocoder '%s'; options are: %s" %
             (service, SERVICE_TO_GEOCODER.keys())
         )
-
--- a/geopy/geocoders/pickpoint.py
+++ b/geopy/geocoders/pickpoint.py
@@ -7,7 +7,8 @@ from geopy.geocoders import Nominatim
 
 from geopy.geocoders.base import (
     DEFAULT_FORMAT_STRING,
-    DEFAULT_TIMEOUT
+    DEFAULT_TIMEOUT,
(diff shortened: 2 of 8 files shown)


geopy/geocoders/pickpoint.py
```diff
@@ -0,0 +1,74 @@
...
+        :param int timeout: Time, in seconds, to wait for the geocoding service
+            to respond before raising a :class:`geopy.exc.GeocoderTimedOut`
+            exception.
+        :param dict proxies: If specified, routes this geocoder's requests
+            through the specified proxy. E.g., {"https": "192.0.2.0"}. For
+            more information, see documentation on
+            :class:`urllib2.ProxyHandler`.
+        """
+
+        Nominatim.__init__(
```
KostyaEsmukov: Any reason to not use `super(PickPoint, self).__init__(` here? The base class `Geocoder` is a new-style class (it extends `object`), so it should be fine. Also all other geocoders call constructor this way.

I would also suggest to use keyword arguments instead of positional ones, because this is really a large set of them.

geopy/geocoders/pickpoint.py
```diff
@@ -0,0 +1,74 @@
...
+            self,
+            format_string,
+            view_box,
+            country_bias,
+            timeout,
+            proxies,
+            'api.pickpoint.io',
+            'https',
```
KostyaEsmukov: `Nominatim` class has `domain` and `scheme` args, so these should be used there too instead of hardcoding the values.

geopy/geocoders/__init__.py
```diff
@@ -105,6 +105,7 @@
 from geopy.geocoders.opencage import OpenCage
 from geopy.geocoders.openmapquest import OpenMapQuest
 from geopy.geocoders.osm import Nominatim
+from geopy.geocoders.pickpoint import PickPoint
```
KostyaEsmukov: I'm going to make a checklist for adding new geocoder classes next days (I'll attach the link when it's done), but before it's there, here's a couple of things definitely missing here:

1. `PickPoint` should be listed in the `__all__`  tuple [above in this file](https://github.com/geopy/geopy/blob/40633f53ce4b85e8e6a7cb0d62d9f0b35864050f/geopy/geocoders/__init__.py#L74)
2. Same for `SERVICE_TO_GEOCODER` dict ([below in this file](https://github.com/geopy/geopy/blob/40633f53ce4b85e8e6a7cb0d62d9f0b35864050f/geopy/geocoders/__init__.py#L125))
3. Should have own test module (done) which should be referenced in [`geopy/test/geocoders/__init__.py`](https://github.com/geopy/geopy/blob/40633f53ce4b85e8e6a7cb0d62d9f0b35864050f/test/geocoders/__init__.py) (missing)
4. In case an API KEY is required, it should be listed in [`geopy/test/geocoders/util.py` (see the top of the file)](https://github.com/geopy/geopy/blob/40633f53ce4b85e8e6a7cb0d62d9f0b35864050f/test/geocoders/util.py#L15) and the test class should be guarded with a unittest decorator (see [`test/geocoders/what3words.py`](https://github.com/geopy/geopy/blob/40633f53ce4b85e8e6a7cb0d62d9f0b35864050f/test/geocoders/what3words.py#L18) for example).

test/geocoders/pickpoint.py
```diff
@@ -0,0 +1,180 @@
+from geopy.compat import u
+from geopy.point import Point
+from geopy.geocoders import PickPoint
+from test.geocoders.util import GeocoderTestBase
+
+
+class PickPointTestCase(GeocoderTestBase):
+
+    delta = 0.04
+    API_KEY="4Xbbg6q9Tzms7ZyrrtvJ"
```
KostyaEsmukov: Please use an ENV variable instead

cylon-v:
@KostyaEsmukov, thanks for the review! I will update the PR soon.

cylon-v:
Hey @KostyaEsmukov, I updated my PR, please feel free to review changed version.

KostyaEsmukov:
I've just pushed a couple of cosmetic changes to your branch. I'm sorry for not asking you first: I just want to speed up the 1.13 version release.

Merging now, thank you for this!

PR #260, opened 2018-01-30
Change default coordinates BD09 to standard WGS84

ArkmanKing:
This a bugfix, because GPS coordinates don't work with BD09.
Code has been tested with some beijing GPS lat/lon pairs.

### DIFF
--- a/geopy/geocoders/baidu.py
+++ b/geopy/geocoders/baidu.py
@@ -118,6 +118,7 @@ class Baidu(Geocoder):
             'ak': self.api_key,
             'output': 'json',
             'location': self._coerce_point_to_string(query),
+            'coordtype': 'wgs84ll',
         }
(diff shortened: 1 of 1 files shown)


ArkmanKing:
Don't merge this commit into 1.15. See below.

As described in http://lbsyun.baidu.com/index.php?title=webapi/guide/webservice-geocoding-abroad
It said that the default is BD09LL for coordtype parameter in requests. 

But baidu works fine in china with default BD09ll coordinate, because GPS84 coordinate is not allowed. I will prepare another pull request later.

KostyaEsmukov:
Thanks for letting me know. I suggest to make the coordinates type a geocoder's option, so people could choose that fits them, if that makes sense. Looking forward to that next PR of yours!

Issue #127, opened 2015-05-26
bounds breaks opencage geocoder


ehowland:
This test code fragment works, yielding a geolocation (43.074718, -89.386238)

```
from geopy.geocoders import OpenCage
query = "100 State St, Madison WI 53711, USA";
geolocator = OpenCage(key, timeout=3600)
location = geolocator.geocode(query,exactly_one=True,
#                              bounds="42.8448769,-89.8394169,43.2940373,-89.0094259",
                              country="USA"
                              )
print (location.address)
```

but if I uncomment the bounds line the search comes up empty and the subsequent attempt to use the search result errors out:

```
Traceback (most recent call last):
  File "test_geopy.py", line 39, in <module>
    print (location.address)
AttributeError: 'NoneType' object has no attribute 'address'

```

mtmail:
The documentation wasn't clear in which order the coordinates are expected. Fix in https://github.com/geopy/geopy/pull/264

PR #264, opened 2018-03-12
clarify OpenCage's bounds parameter

mtmail:
Fixes https://github.com/geopy/geopy/issues/127
Documentation says "south-west and north-east corners" which is easy to read as lat,lon,lat,lon while the OpenCage API expects lon,lat,lon,lat.

### DIFF
--- a/geopy/geocoders/opencage.py
+++ b/geopy/geocoders/opencage.py
@@ -90,9 +90,10 @@ class OpenCage(Geocoder):
             but will not restrict the possible results to the supplied
             region. The bounds parameter should be specified as 4
             coordinate points forming the south-west and north-east
-            corners of a bounding box. For example,
-            `bounds=-0.563160,51.280430,0.278970,51.683979`.
-            (min lon, min lat, max lon, max lat)
+            corners of a bounding box. The order of the coordinates is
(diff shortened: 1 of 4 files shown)


geopy/geocoders/opencage.py
```diff
@@ -88,6 +88,7 @@ def geocode(
             coordinate points forming the south-west and north-east
             corners of a bounding box. For example,
             `bounds=-0.563160,51.280430,0.278970,51.683979`.
+            (min lon, min lat, max lon, max lat)
```
KostyaEsmukov: Could you please add some context here?

Something like this will do, IMO:
```
             The order of coordinates is: `(min lon, min lat, max lon, max lat)`.
```

mtmail: Of course, changed now. (it was a copy&paste from https://geocoder.opencagedata.com/api#forward). Thanks a lot for taking over as maintainer!

mtmail:
Messed up my `git rebase` after `README.md` got renamed to `README.rst`. I created a new cleaner PR in https://github.com/geopy/geopy/pull/265

KostyaEsmukov:
You could've used force push to the same branch w/o recreating PR, but that's OK. 

Thank you, the new PR has been merged.

PR #266, opened 2018-03-17
Replace nosetest with pytest

KostyaEsmukov:
- [x] Blocked by #144 (`test/test_distance.py` is significantly rewritten here, and I'm not willing to cause additional roadblocks for #144).
- [x] Add usage docs


KostyaEsmukov:
OK, I believe it's ready. Let's throw this in and see how it goes.

As for the migration from nose, I'd say the following:
- There's no need to import tests in `__init__.py` files anymore. pytest discovers them automatically.
- There's no need to add doc strings to each test method anymore. They're not printed anyway.
- To start using pytest, just follow the steps at CONTRIBUTING.md.

Issue #245, opened 2017-06-23
UnboundLocalError: local variable 'cos_sq_alpha' referenced before assignment


1kastner:
First of all thanks for the great contribution! Most of the times the library serves well, but recently I started to get the following kinds of errors:

    Traceback (most recent call last):
      ...
      File "evaluate_all_methods.py", line 226, in score_algorithm
        point = geopy.Point(lat, lon)
      File "...\lib\site-packages\geopy\distance.py", line 322, in __init__
        super(vincenty, self).__init__(*args, **kwargs)
      File "...\lib\site-packages\geopy\distance.py", line 115, in __init__
        kilometers += self.measure(a, b)
      File "...\lib\site-packages\geopy\distance.py", line 414, in measure
        u_sq = cos_sq_alpha * (major ** 2 - minor ** 2) / minor ** 2
    UnboundLocalError: local variable 'cos_sq_alpha' referenced before assignment

My code is quite complex and it happens here and then while processing bulk requests. Because of that I can not easily break it down to the place where it fails.

1kastner:
One thing I found in the code: when the distance is too short, the while-loop is never entered so that cos_sq_alpha is not set. Especially the algorithm does not handle np.nan values well.

1kastner:
I would suggest a check of the variables before the distance is calculated. As a numpy import for checking for NaNs would be an overkill, maybe a smaller and greater zero can be checked, like:

    is_nan = lambda x: not (x > 0 or x < 0)

That way the library will turn more user friendly.

KostyaEsmukov:
Should be fixed by 15ebe15. Feel free to reopen if the problem persists.

Also linking #144 which actually contained the fix (but without tests).

KostyaEsmukov:
@1kastner I like the input checking suggestion. I'd gladly accept a PR that does this.

KostyaEsmukov:
#272 would introduce some input validation

Issue #242, opened 2017-04-27
UnboundLocalError at vincenty distance


kataev:
```
import sys
print(sys.version_info)
print(geopy.__version__)
geopy.distance.vincenty((np.NaN, np.NaN), (1,1))

sys.version_info(major=3, minor=6, micro=0, releaselevel='final', serial=0)
1.11.0
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-87-a808b291b901> in <module>()
      2 print(sys.version_info)
      3 print(geopy.__version__)
----> 4 geopy.distance.vincenty((np.NaN, np.NaN), (1,1))

/Users/pk/miniconda3/lib/python3.6/site-packages/geopy/distance.py in __init__(self, *args, **kwargs)
    320         self.iterations = kwargs.pop('iterations', 20)
    321         major, minor, f = self.ELLIPSOID # pylint: disable=W0612
--> 322         super(vincenty, self).__init__(*args, **kwargs)
    323 
    324     def set_ellipsoid(self, ellipsoid):

/Users/pk/miniconda3/lib/python3.6/site-packages/geopy/distance.py in __init__(self, *args, **kwargs)
    113         elif len(args) > 1:
    114             for a, b in util.pairwise(args):
--> 115                 kilometers += self.measure(a, b)
    116 
    117         kilometers += units.kilometers(**kwargs)

/Users/pk/miniconda3/lib/python3.6/site-packages/geopy/distance.py in measure(self, a, b)
    412             raise ValueError("Vincenty formula failed to converge!")
    413 
--> 414         u_sq = cos_sq_alpha * (major ** 2 - minor ** 2) / minor ** 2
    415 
    416         A = 1 + u_sq / 16384. * (

UnboundLocalError: local variable 'cos_sq_alpha' referenced before assignment
```

cffk:
Perhaps it's worth revisiting pull request #144.  Surely geopy should
offer a method for computing geodesic distances that "just works".

cffk:
The same error can be triggered without NaNs with
~~~
geopy.distance.vincenty((0,-180),(0,180))
~~~

KostyaEsmukov:
Closing this, as I'm unable to reproduce the issue. Added the tests in 0ea5e8d and they currently pass. Feel free to reopen if anyone finds another test which fails.

cffk:
@KostyaEsmukov You fixed this with commit 15ebe153b.

[...]

PR #272, opened 2018-04-01
Add input validation warnings to Point

KostyaEsmukov:
This PR adds coordinates validation and issues warnings when they're invalid (or error-prone).

TODO:
- [x] maybe don't restrict longitude range? See discussion below
- [x] add warnings to the Point constructor https://github.com/geopy/geopy/issues/242#issuecomment-377886172
- [x] error messages and comments should be grammar-checked and refined a bit

~~Regarding performance degradation – it's \~1.2 times slower with Points and \~1.9 times slower with tuples for me. IMO it's tolerable.~~
 
Related discussions: #245 #242 #202

### DIFF
--- a/geopy/compat.py
+++ b/geopy/compat.py
@@ -36,6 +36,16 @@ if py3k:
 else:
     cmp = cmp  # builtin in py2
 
+
+if py3k:
+    from math import isfinite
+else:
+    from math import isinf, isnan
+
+    def isfinite(x):
+        return not isinf(x) and not isnan(x)
+
+
 if py3k: # pragma: no cover
     from urllib.parse import (urlencode, quote, # pylint: disable=W0611,F0401,W0611,E0611
                               urlparse, parse_qs)
--- a/geopy/point.py
+++ b/geopy/point.py
@@ -5,6 +5,7 @@
 
 import collections
 import re
+import warnings
 from math import fmod
 from itertools import islice
 from geopy import util, units
@@ -15,7 +16,7 @@ from geopy.format import (
     format_degrees,
     format_distance,
 )
-from geopy.compat import string_compare
+from geopy.compat import string_compare, isfinite
 
 
 POINT_PATTERN = re.compile(r"""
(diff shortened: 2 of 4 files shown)


cffk:
I think that longitudes in the range `[-360,360]` should be accepted without complaint.  In some applications even wider ranges for longitude make sense, e.g., when you're keeping track of how many times you've encircled the globe.

KostyaEsmukov:
But `[-360,360]` overlaps the circle two times, which seems ambiguous. How is that different from `[-4pi,4pi]`, which overlaps the circle four times? In that case I think it's better to not restrict longitude range at all, if we tend to think that a longitude outside `[-180,180]` is not a mistake. Thanks for pointing this out, I'll reason about it.

KostyaEsmukov:
I moved all the warnings from Distance to Point, which allowed to simplify the validation code significantly. For the reference, the previous commit was 58a37a839988d4db2413f38881f36b1d9001c637.

I also decided to not complain about longitude normalization, because I believe there's nothing error-prone with that (unlike latitude).

I'll leave it open for a day to reason a bit about this all, but I guess it's ready as-is. I would be glad if someone could review (especially the warning messages, as I'm not so confident with English).

KostyaEsmukov:
Ok, I refined one message a little. Now the warnings look like this:

```
>>> from geopy.distance import distance
>>> distance((170, 20), (170, 10))
[...]/geopy/geopy/point.py:85: UserWarning: Latitude has been normalized, because its absolute value was more than 90. This is probably not what was meant, because the normalized value is on a different pole. If you pass coordinates as positional args, please make sure that the order is (latitude, longitude) or (y, x) in Cartesian terms.  This will cause a ValueError exception in the future versions of geopy.
  UserWarning)
Distance(1096.3515545690054)
>>> distance(20, 170)
[...]/geopy/geopy/point.py:184: UserWarning: A single number has been passed to the Point constructor. This is probably a mistake, because constructing a Point with just a latitude seems senseless. If this is exactly what was meant, then pass the zero longitude explicitly to get rid of this warning.
  UserWarning)
Distance(3318.2210874060056)
>>> distance((20, float("nan")), (10, 170))
[...]/geopy/geopy/point.py:74: UserWarning: Point coordinates should be finite. NaN or inf has been passed as a coordinate.  This will cause a ValueError exception in the future versions of geopy.
  UserWarning)
Distance(nan)
```

Well, I hope it won't break anything. Merging and prepping the `1.13.0` release.

PR #279, opened 2018-04-08
Point: fix unwanted precision loss

KostyaEsmukov:
See discussion at cddf65d. It turns out that replacing `%` with `fmod` alone (which was done in #275) wasn't enough to fix unwanted precision loss on Point coordinates normalization.

This PR (hopefully) fixes that issue.

@cffk could you take a look, please?

cc @svalee (just FYI)

### DIFF
--- a/geopy/point.py
+++ b/geopy/point.py
@@ -47,6 +47,19 @@ POINT_PATTERN = re.compile(r"""
 }, re.X)
 
 
+def _normalize_angle(x, limit):
+    """
+    Normalize angle `x` to be within `[-limit; limit)` range.
+    """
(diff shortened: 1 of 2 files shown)


cffk:
This looks good.  (I'm still concerned about turning `lat = 90 + eps` to `lat = -90 + eps`; but that's a discussion for another day...)

KostyaEsmukov:
> (I'm still concerned about turning lat = 90 + eps to lat = -90 + eps; but that's a discussion for another day...)

Absolutely, this is still an issue. I'll add a warning for that in #272.

Thanks for reviewing!

Issue #170, opened 2016-01-20
In OSM if exactly_one=True expect to have in query "limit=1" to limit http transport


adamailru:
I have the code below
    
   >>> print(request.POST)
   <QueryDict: {'city': ['Chisinau'], 'street': ['str. Alecu Russo'], 'language': ['ro'], 'country': ['Moldova'], 'csrfmiddlewaretoken': ['RMScTXnPa40Lu6fxdLcPgqrTti7BHONn'], 'limit': ['1']}>
   >>> location = Nominatim().geocode(request.POST, exactly_one=True)

I'm expecting to have something
=========================
     
     Nominatim.geocode: https://nominatim.openstreetmap.org/search?country=Moldova&city=Chisinau&street=str.+Igor+Vieru&format=json&limit=1
                                                                                 ^^^^^^
But I got something like this 
======================

    Nominatim.geocode: https://nominatim.openstreetmap.org/search?city=Chisinau&country=Moldova&street=str.+Alecu+Russo&format=json

*no limit!*

Is this a missed feature or it should be alternative approach to minimize http transport ?

KostyaEsmukov:
`limit` kwarg is just not implemented in Nominatim geocoder yet. This is certainly a valid feature request. Photon geocoder [implements this](https://github.com/geopy/geopy/blob/4c8cbf056175b27d87242c2b96c0ca0178c91907/geopy/geocoders/photon.py#L117), so it can be taken as an inspiration. PR is welcome.

PR #281, opened 2018-04-09
Implement Nominatim geocode result limit

Serphentas:
This is just a quick fix for #170.

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -114,6 +114,7 @@ class Nominatim(Geocoder):
             self,
             query,
             exactly_one=True,
+            limit=None,
             timeout=None,
             addressdetails=False,
             language=False,
@@ -179,6 +180,15 @@ class Nominatim(Geocoder):
         params.update({
             'format': 'json'
         })
+
+        if limit:
+            limit = int(limit)
+            if limit < 1:
+                raise ValueError("Limit cannot be less than 1")
+            exactly_one = limit == 1
+            params['limit'] = int(limit)
+        elif exactly_one:
+            params['limit'] = 1
 
         # `viewbox` apparently replaces `view_box`
         if self.view_box:


geopy/geocoders/osm.py
```diff
@@ -180,6 +181,15 @@ def geocode(
             'format': 'json'
         })
 
+        if limit:
+            limit = int(limit)
+            if limit < 1:
+                raise ValueError("Limit cannot be less than 1")
+            exactly_one = limit == 1
```
KostyaEsmukov: `exactly_one` affects the return type of the function call (a single location vs a list of locations), so I believe that in case of conflicting `limit` and `exactly_one` values the latter should take precedence.

What do you think of removing this line and turning the following `elif` to a simple `if`, which would force `limit=1` for `exactly_one=True`?

geopy/geocoders/osm.py
```diff
@@ -114,6 +114,7 @@ def geocode(
             self,
             query,
             exactly_one=True,
+            limit=None,
```
KostyaEsmukov: Could you please make this arg the latest one? [The base method](https://github.com/geopy/geopy/blob/d3e076a8ce0cf47667af36d28c40d24039ac516f/geopy/geocoders/base.py#L205) has `timeout` as its last arg and this order should be preserved to not cause a strange behaviour when `geocode` is called using positional args.

Also it would be nice to add a docstring here (with `versionadded:: 1.13.0`). You can borrow one [from Photon, for example](https://github.com/geopy/geopy/blob/d3e076a8ce0cf47667af36d28c40d24039ac516f/geopy/geocoders/photon.py#L101).

Serphentas: You mean limit should be placed after timeout, is that right ?

KostyaEsmukov: After `timeout` or as the last one (after `geometry`) – both options are fine.

geopy/geocoders/osm.py
Serphentas: Just to be sure, is this what you have in mind ?

```python
if exactly_one:
  params['limit'] = 1
elif limit:
  limit = int(limit)
  if limit < 1:
    raise ValueError("Limit cannot be less than 1")
  params['limit'] = limit
```

If I write that, given that `exactly_one` is true by default, we will always have that `limit=1`. Also, the result will always be a single location result (and not a list, even if a limit is given). To change this (i.e. have a limit other than one **and** a list of results), we would have to explicitly set `exactly_one` to false in addition to passing a limit, which is a bit cumbersome in my opinion.

I initially wanted to raise an exception if both exactly_one and limit were set, but as said above such a case is not ideal. That's why I made it so that if limit is given, the value of exactly_one is changed and determined by `exactly_one = limit == 1`.

What are your thoughts on this ?

geopy/geocoders/osm.py
Serphentas: Done in cd9eea9

geopy/geocoders/osm.py
KostyaEsmukov: I was trying to describe something like this:
```python
        if limit:
            limit = int(limit)
            if limit < 1:
                raise ValueError("Limit cannot be less than 1")
            params['limit'] = limit
        if exactly_one:
            params['limit'] = 1
```

Your snippet does the same (except that in your case a ValueError won't be thrown for `exactly_one=True, limit=-1`). Both are fine IMO, pick the one that you prefer.

> To change this (i.e. have a limit other than one and a list of results), we would have to explicitly set exactly_one to false in addition to passing a limit, which is a bit cumbersome in my opinion.

It's the current convention to explicitly set `exactly_one` to False for getting a list. I kinda agree that it looks cumbersome to specify both, but violating that convention doesn't seem reasonable to me.

`exactly_one` is implemented by all geocoders. Basically it means what kind of result does the user expect from this geocoding call. It's universal across all geocoder classes. `limit` is just an optional arg of this specific geocoder which allows to tell the backend a *desired* amount of results. But `limit` in no way controls the real number of locations in a result: it can be less, it can be more. It's up to the backend to decide.

So the idea is that `exactly_one` always takes precedence over `limit`.

Thereby I think there's no need to raise an exception if both are set – from my point of view it seems valid (except the `exactly_one=True` and `limit != 1` case, which can be silently fixed to `limit = 1` without being ambiguous to the API user, given that it's `exactly_one` arg which controls the response type rather than `limit`).

Does that seem reasonable to you?

KostyaEsmukov:
BTW I've just realized that `limit` should also be added to the `reverse` function in the same fashion. Could you do that, please?

Serphentas:
I actually implemented the feature for `reverse` too but then I realized that Nominatim has no limit option for reverse geocoding (see [API parameters](https://wiki.openstreetmap.org/wiki/Nominatim#Parameters_2)). I even tried to send [requests with a limit](https://nominatim.openstreetmap.org/reverse?format=json&lat=52.5487429714954&lon=-1.81602098644987&limit=3) but I always got one location only, so I assume limits don't apply in this case.

KostyaEsmukov:
Oh, I see. Yup, sure, in this case `reverse` should not accept a `limit` kwarg.

geopy/geocoders/osm.py
Serphentas: Put this way it is wiser to continue with the current convention you mentioned. As such, it should be made clear in the documentation how the limit parameter is handled.

Doing changes now.

Serphentas: Done in 05651b5

Serphentas:
Thanks !

PR #282, opened 2018-04-13
Added 2 helper functions lonlat and latlon, added documentation for each

### LINKED ISSUE #202: Distance calculation takes (y, x) instead of (x, y)
So the issue is that distance caluclation (`vincenty` and `great_circle`) take arguments in the form of 
 `((y, x), (y, x))` instead of `((x, y), (x, y))`. The convention followed by Geopy is consistent with the idea of `(Latitude, Longitude)`, but is not conversant with Shapely or Fiona's idea of the same. Shapely and Fiona will return `(x, y)`, or `(Longitude, Latitude)`, and I think Geopy should accept that format as default. Perhaps have a Boolean switch that takes the other format. For example, the following pseudocode:

```
def vincenty(p1, p2, xy=True):
    if xy:
        p1 = reversed(p1)
        p2 = reversed(p2)
    # And the rest of the algorithm as usual.
```

IMHO, this would save a considerable amount of headache. Most answers on Stack Exchange are also oblivious to this (eg: http://gis.stackexchange.com/questions/4022/looking-for-a-pythonic-way-to-calculate-the-length-of-a-wkt-linestring).

svalee:
this commit fixes #202


geopy/distance.py
```diff
@@ -86,7 +86,6 @@
     >>> _, pa = g.geocode('Palo Alto, CA')
     >>> print((d(ne, cl) + d(cl, wa) + d(wa, pa)).miles)
     3277.30439191
-
```
KostyaEsmukov: Could you please revert this change?

geopy/distance.py
```diff
@@ -120,6 +119,39 @@
...
+
+
+def latlon(y, x, z=0):
+    """
+    Helper function that can be used for more explicit and obvious calling of the distance methods
+    >>> from geopy.distance import latlon, distance
+    >>> newport_ri = (41.49008, -71.312796)
+    >>> cleveland_oh = (41.499498, -81.695391)
+    >>> print(distance(latlon(*newport_ri), latlon(*cleveland_oh)).miles)
+    538.3904453677203
```
KostyaEsmukov: Parameters (below) must be separated from description (above) with one empty line:

```
"""
... description ...
  <-- empty newline
:param ...
"""
```

geopy/distance.py
```diff
@@ -120,6 +119,39 @@
...
+    :param z: altitude
+    :return: Point(latitude, longitude, altitude)
+    """
+    return Point(y, x, z)
+
+
+def latlon(y, x, z=0):
+    """
+    Helper function that can be used for more explicit and obvious calling of the distance methods
+    >>> from geopy.distance import latlon, distance
```
KostyaEsmukov: I believe that code examples must be indented to be rendered correctly, though I'm not sure. Anyway, all the examples in this file *are* indented, so could you please do that there too?

geopy/distance.py
```diff
@@ -120,6 +119,39 @@
...
+    >>> newport_ri = (41.49008, -71.312796)
+    >>> cleveland_oh = (41.499498, -81.695391)
```
KostyaEsmukov: IMO it might be not easy to comprehend that this is an y-x order, but this is crucial for understanding the intent of this function. Maybe a more explanatory variable name would help?

`newport_ri_yx = (41.49008, -71.312796)`

geopy/distance.py
```diff
@@ -120,6 +119,39 @@
     'GRS-67':        (6378.1600, 6356.774719, 1 / 298.25)
 }
 
+
+def lonlat(x, y, z=0):
+    """
+    Helper function that can be used for more explicit and obvious calling of the distance methods
```
KostyaEsmukov: I don't think that `obvious` is a proper term for documentation :-)

This description should introduce the reader to the problem: `distance` accepts coordinates in `(y, x)`/`(lat, lon)` order, while some other libraries/systems might use `(x, y)`/`(lon, lat)` instead.

And then suggest a solution: using this function, which accepts coordinates in `(x, y, [z])` order and returns a Point instance for the `distance`.

geopy/distance.py
```diff
@@ -120,6 +119,39 @@
...
+    :param y: latitude
+    :param z: altitude
+    :return: Point(latitude, longitude, altitude)
+    """
+    return Point(y, x, z)
+
+
+def latlon(y, x, z=0):
+    """
+    Helper function that can be used for more explicit and obvious calling of the distance methods
```
KostyaEsmukov: This function is very different from the previous one, so should be the description. I believe it will be confusing to the reader to see a function which transforms `(y, x, [z])` to the same `(y, x, [z])`.

Actually I'm starting to think that instead of introducing this function, it would probably be better to simply encourage the use of the Point constructor instead. What do you think?

svalee:
Hey, it's been a while, I'm verry sorry. I got sick and then a lot of work appeared suddenly.
Anyway, so should I delete latlon function?
And did build failed? I don't get it, I didn't change anything other than docs.

geopy/distance.py
```diff
@@ -122,12 +123,17 @@
 
 def lonlat(x, y, z=0):
     """
-    Helper function that can be used for more explicit and obvious calling of the distance methods
+    distance accepts coordinates in (y, x)/(lat, lon) order, while some other libraries/systems might use
```
KostyaEsmukov: 1. A sentence should start with a capital letter :). But in this case `distance` is a term (a class reference, more specifically), not a plain word. See the docs around for examples on how `distance` is formatted there.
2. Please try to keep the line lengths close to 80-ish.
3. Coordinate pairs should also be formatted as terms (with backticks, I guess), otherwise they might be split by word-wrapping.

geopy/distance.py
```diff
@@ -122,12 +123,17 @@
 
 def lonlat(x, y, z=0):
     """
-    Helper function that can be used for more explicit and obvious calling of the distance methods
+    distance accepts coordinates in (y, x)/(lat, lon) order, while some other libraries/systems might use 
+    (x, y)/(lon, lat).
+    This helper function introduced here for solving this problem. It accepts coordinates in a form of (x, y)/(lon, lat)
+    and returns Point that is safe to use with distance
```
KostyaEsmukov: 1. A sentence should end with a dot (or some other relevant punctuation mark).
2. It seems to me that these two sentences are intended to be in their own paragraph. Paragraphs should be separated with one empty line, which is missing there.
3. These two sentences don't sound natural to me. Let me propose the following phrasing:

        This function provides a convenient way to convert coordinates of 
        the ``(x, y)/(lon, lat)`` format to a :class:`geopy.point.Point` instance.

        Example::

svalee:
Hey again, thanks for positive feedback, I am doing my best, but it's actually pretty hard to write meaningful docs, so thank you very much for your guidance and patience!

geopy/distance.py
```diff
@@ -86,7 +86,7 @@
     >>> _, pa = g.geocode('Palo Alto, CA')
     >>> print((d(ne, cl) + d(cl, wa) + d(wa, pa)).miles)
     3277.30439191
-
+
```
KostyaEsmukov: This change should be reverted

geopy/distance.py
```diff
@@ -120,6 +120,31 @@
...
+    
+    >>> from geopy.distance import lonlat, distance
+    >>> newport_ri = (-71.312796, 41.49008)
+    >>> cleveland_oh = (-81.695391, 41.499498)
+    >>> print(distance(lonlat(*newport_ri), lonlat(*cleveland_oh)).miles)
+    538.3904453677203
+    
+    :param x: longitude 
+    :param y: latitude
+    :param z: altitude
```
KostyaEsmukov: `:param z: (optional) altitude`

geopy/distance.py
```diff
@@ -120,6 +120,31 @@
...
+    """
+    `geopy.distance.distance` accepts coordinates in ``(y, x)/(lat, lon)`` order, while some other libraries/systems
+    might use ``(x, y)/(lon, lat)``.
+
+    This function provides a convenient way to convert coordinates of the ``(x, y)/(lon, lat)`` format to
+    a :class:`geopy.point.Point` instance.
+
+    Example::
+    
+    >>> from geopy.distance import lonlat, distance
```
KostyaEsmukov: Code blocks should be indented.

Like this: https://github.com/svalee/geopy/blob/f4d0b509dccc1f8a937e1ae2416066bcf611f08b/geopy/distance.py#L289

geopy/distance.py
```diff
@@ -120,6 +120,31 @@
...
+    >>> newport_ri = (-71.312796, 41.49008)
```
KostyaEsmukov: `newport_ri_xy` and `cleveland_oh_xy`

KostyaEsmukov:
We seem to have forgotten about the tests! Added in fdfaac7.

Also the import in the docs was incorrect: it was

```
.. autofunction:: geopy.lonlat
```
... which yielded an error when building the docs, because lonlat is not exported from geopy. It should've been this:

```
.. autofunction:: geopy.distance.lonlat
```

I guess it was just a typo.

PR #298, opened 2018-05-16
Baidu: signature authentication support

jinzhc:
changes for baidu geocoder & pytest
For signature authentication, see
http://lbsyun.baidu.com/index.php?title=lbscloud/api/appendix

### DIFF
--- a/geopy/geocoders/baidu.py
+++ b/geopy/geocoders/baidu.py
@@ -2,7 +2,7 @@
 :class:`.Baidu` is the Baidu Maps geocoder.
 """
 
-from geopy.compat import urlencode, quote, quote_plus
+from geopy.compat import quote, quote_plus
 from geopy.exc import (
     GeocoderServiceError,
     GeocoderAuthenticationFailure,
@@ -123,9 +123,9 @@ class Baidu(Geocoder):
         params = {
             'ak': self.api_key,
             'output': 'json',
-            'address': self.format_string % query,
+            'address': quote(self.format_string % query, safe="!*'();:@&=+$,/?#[]"),
         }
-        params = urlencode(params, safe="!*'();:@&=+$,/?#[]", quote_via=quote)
+        params = "&".join(["{}={}".format(k, v) for k, v in params.items()])
 
         if self.security_key is None:
             url = "?".join((self.api, params))
@@ -163,9 +163,12 @@ class Baidu(Geocoder):
         params = {
             'ak': self.api_key,
             'output': 'json',
-            'location': self._coerce_point_to_string(query),
+            'location': quote(
+                self._coerce_point_to_string(query),
+                safe="!*'();:@&=+$,/?#[]"
+                ),
         }
-        params = urlencode(params, safe="!*'();:@&=+$,/?#[]", quote_via=quote)
+        params = "&".join(["{}={}".format(k, v) for k, v in params.items()])
 
         if self.security_key is None:
             url = "?".join((self.api, params))
--- a/geopy/geocoders/baidu.py
+++ b/geopy/geocoders/baidu.py
@@ -83,8 +83,8 @@ class Baidu(Geocoder):
             ssl_context=ssl_context,
         )
         self.api_key = api_key
-        self.api = '%s://api.map.baidu.com/geocoder/v2/' % self.scheme
-        self.api_path = self.api.split('baidu.com')[1]
+        self.api_path = '/geocoder/v2/'
+        self.api = '%s://api.map.baidu.com%s' % (self.scheme, self.api_path)
         self.security_key = security_key
 
     @staticmethod
@@ -255,11 +255,11 @@ class Baidu(Geocoder):
             raise GeocoderQueryError(
                 'IP/SN/SCODE/REFERER Illegal:'
             )
-        elif status in range(200, 300):
+        elif 200 <= status <= 300:
             raise GeocoderQueryError(
                 'Has No Privilleges.'
             )
-        elif status in range(300, 400):
+        elif 301 <= status <= 402:
             raise GeocoderQuotaExceeded(
                 'Quota Error.'
             )
--- a/geopy/geocoders/baidu.py
+++ b/geopy/geocoders/baidu.py
@@ -4,6 +4,7 @@
 
 from geopy.compat import urlencode, quote_plus
 from geopy.exc import (
+    GeocoderServiceError,
     GeocoderAuthenticationFailure,
     GeocoderQueryError,
     GeocoderQuotaExceeded,
@@ -228,7 +229,7 @@ class Baidu(Geocoder):
             # When there are no results, just return.
             return
         if status == 1:
-            raise GeocoderQueryError(
+            raise GeocoderServiceError(
                 'Internal server error.'
             )
         elif status == 2:
@@ -248,16 +249,24 @@ class Baidu(Geocoder):
                 'AK Illegal or Not Exist.'
             )
         elif status == 101:
-            raise GeocoderQueryError(
-                'Your request was denied.'
+            raise GeocoderAuthenticationFailure(
+                'No AK'
             )
         elif status == 102:
-            raise GeocoderQueryError(
-                'IP/SN/SCODE/REFERER Illegal:'
+            raise GeocoderAuthenticationFailure(
+                'MCODE Error'
+            )
+        elif status == 200:
+            raise GeocoderAuthenticationFailure(
+                'Invalid AK'
+            )
+        elif status == 211:
+            raise GeocoderAuthenticationFailure(
+                'Invalid SK'
             )
         elif 200 <= status <= 300:
-            raise GeocoderQueryError(
-                'Has No Privilleges.'
+            raise GeocoderAuthenticationFailure(
+                'Authentication Failure'
             )
         elif 301 <= status <= 402:
             raise GeocoderQuotaExceeded(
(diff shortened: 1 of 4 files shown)


ArkmanKing:
I have done the pytest with my api_keys with/without security key

geopy/geocoders/baidu.py
```diff
@@ -78,6 +84,8 @@ def __init__(
         )
         self.api_key = api_key
         self.api = '%s://api.map.baidu.com/geocoder/v2/' % self.scheme
+        self.api_path = self.api.split('baidu.com')[1]
```
KostyaEsmukov: Maybe that would be better?
```
self.api_path = '/geocoder/v2/'
self.api = '%s://api.map.baidu.com%s' % (self.scheme, self.api_path)
```

geopy/geocoders/baidu.py
```diff
@@ -205,44 +224,55 @@ def _check_status(status):
...
             raise GeocoderQueryError(
                 'Your request was denied.'
             )
-        elif status == '102':
+        elif status == 102:
             raise GeocoderQueryError(
                 'IP/SN/SCODE/REFERER Illegal:'
             )
-        elif status == '2xx':
+        elif status in range(200, 300):
```
KostyaEsmukov: In Python2 `range` is a list, which seems inefficient.
What about this?

```
elif 200 <= status <= 300:
```

test/geocoders/baidu.py
```diff
@@ -27,6 +27,7 @@ class BaiduTestCase(GeocoderTestBase):
     def setUpClass(cls):
         cls.geocoder = Baidu(
             api_key=env['BAIDU_KEY'],
+            security_key=env['BAIDU_SEC_KEY'],
```
KostyaEsmukov: I suggest to create a separate test class instead, which does use the SK, so we could test both setups. I think that having two simple calls (1 geocode and 1 reverse) for this new class would be enough to be sure that SK works.

Also, if I understood correctly, that should be a different `BAIDU_KEY` (for which the SK auth is enabled), shouldn't it? If so, it should have a different ENV name.

geopy/geocoders/baidu.py
```diff
@@ -205,44 +224,55 @@ def _check_status(status):
...
             raise GeocoderQueryError('Unknown error')
+
+    def _url_with_signature(self, quoted_params):
+        """
+        Return the request url with signature.
+        qutoted_params is the escaped str of params in request.
+        """
+
+        # Since quoted_params is escaped by urlencode, don't apply quote twice
+        raw = self.api_path + '?' + quoted_params + self.security_key
```
KostyaEsmukov: According to the docs at http://lbsyun.baidu.com/index.php?title=lbscloud/api/appendix , the `quote(params, safe="/:=&?#+!$,;'@()*[]")` code is used for getting a raw string, while there we have `urlencode(params)`. Wouldn't that cause different hashes for peculiar requests? I think the relevant test cases should be added. For example, a one with a comma: `locator.geocode("beijing, china")`.

jinzhc:
Comments for `status` in http response `page`.

See 状态码定义 (`status` definition in English) section (1st section) in http://lbsyun.baidu.com/index.php?title=lbscloud/api/appendix. `status` has more values than original implementation. The failed signature will have 211 status value, whose inferred type is int. This is why I change `status` checking from `str` to `int` version.

geopy/geocoders/baidu.py
jinzhc: The `quoted_params` in `self._url_with_signature` was given by `urlencode`. `urlencode` has already calls `quote_plus` on parameters internally. Except parameters, `quote` never escape the characters in site/path/key parts, because they are not special characters.

test/geocoders/baidu.py
jinzhc: Yes, you're right. Are BAIDU_KEY_REQUIRES_SK and BAIDU_SEC_KEY ok?

KostyaEsmukov: Yup, that's fine.

KostyaEsmukov:
I'm not sure if I understood your reply regarding `status` correctly. My question was specifically about the types, not about checking for more specific error codes. If `status` in the API response is an int, then it would be never equal to a string (In Python `1 == '1'` yields False). And if it's a string in the response, then, again, `'1' == 1` would still yield False.

I've just checked, API returns ints. It means that the current implementation (I mean, without this patch) doesn't seem to respect the `status` at all. It's not covered by tests too. That's pretty sad :(

geopy/geocoders/baidu.py
jinzhc: Yes, this patch code doesn't match the given sample code in docs. But it works.

jinzhc:
Perhaps the new Invalid AK test case will help. It's hard to cover all failure status tests due to the poor Baidu docs.

jinzhc:
For `unittest.assertRaisesRegex`'s compatibility, will`pytest.assertRaises` be a better option?

test/geocoders/baidu.py
```diff
@@ -5,16 +5,34 @@
...
+    def test_user_agent_custom(self):
+        self.assertEqual(self.geocoder.headers['User-Agent'], 'my_user_agent/1.0')
+
+    def test_invalid_ak(self):
+        if py3k:
+            assert_raise_msg = self.assertRaisesRegex
+        else:
+            assert_raise_msg = self.assertRaisesRegexp
+
+        with assert_raise_msg(GeocoderAuthenticationFailure, 'Invalid AK'):
```
KostyaEsmukov: Another way to do the same without tinkering around compatibility:
```
    with self.assertRaises(GeocoderAuthenticationFailure) as cm:
        ...
    self.assertEqual(str(cm.exception), 'Invalid AK')
```

To me it looks better than `assertRaisesRegex`. What do you think?

Also I'm wondering: does the `self.reverse_run` command ever run? I feel that `self.geocode_run` causes an exception being thrown, and the `self.reverse_run` call is never tested. To fix that, each call should be wrapped in their own `assertRaises`/`assertRaisesRegex` context.

geopy/geocoders/baidu.py
KostyaEsmukov: It sure works, but for some cases it might not work. Could you please add the test with a comma, and fix the code, if that fails?

test/geocoders/baidu.py
jinzhc: Thanks, `self.reverse_run` cannot be reached. I will remove this line afterwards.

geopy/geocoders/baidu.py
jinzhc: It's better to stick to docs. I'm going to  change `urlencode(params)` to `urlencode(params, safe="!*'();:@&=+$,/?#[]", quote_via=quote)`.  Will you accept this change?

jinzhc:
Do you think the comma test is a must?

KostyaEsmukov:
Sorry for the delay.

Yes, a comma test is definitely a must. Do you need any help or a guidance with it? Please don't hesitate to ask. Basically all what's needed here is a simple forward geocoding request, which contains a comma in its query. If signature is wrong, an exception would be thrown and the test would fail, proving that there's a bug in the signature implementation.

The cf9a30e...1009fe2 change is somewhat disturbing to me. I would better to not construct the query string manually by escaping individual parts and joining them with `&` and `=`, but make `urlencode` to do this. 

It's important to distinguish between input for the signature algorithm and resulting URL. The former requires the aforementioned `quote` with a list of safe characters, while the latter should use `urlencode`, which is safer. Unless the web server uses urlencoded data (before decoding) for verifying the signature on the server side instead of decoded and encoded with the `quote`... In this case a simple `urlencode`, just like it was before this change, should work correctly (and this is why I wanted to have a test before an implementation).

geopy/geocoders/baidu.py
KostyaEsmukov: Please add the test first :)

But yeah, that would be better than crafting the query string manually. Even better if all of this is not required and a simple `urlencode` works.

KostyaEsmukov:
Would you mind if I finished this by myself? :)

jinzhc:
Never mind. It's my pleasure.

jinzhc:
@KostyaEsmukov Is new test appropriate?

KostyaEsmukov:
I'm wondering if at one point Baidu changes their search algorithm such as the `!*'();:@&=+$,/?[] %` part of the query would make nothing returned.

Actually the reason why I wanted to take this over is that there clearly is a confusion between composing a raw body for the signature and urlencoding the params for url. As I said in my previous comment https://github.com/geopy/geopy/pull/298#issuecomment-394734227 , urlencode should be used for the latter, instead of manual string construction. But for the raw body, apparently, the string must be constructed manually (unless, as said before, their backend uses urlencoded data for signature verification, where this `safe` list of characters is actually useless).

It's just easier for me to finish this by myself than trying to explain this in text. Sorry for that. You've done a great job, a little is left. Is that okay if I finish this by myself? 😉Of course you'd still be credited as a contributor and would get a green box. 😀

jinzhc:
It's ok you take it over and go for final shot. :-)

KostyaEsmukov:
Opened #306 which supersedes this PR.

So it looks like the `safe` list of characters is unnecessary at all! The SN is valid as long as the urlencoded string matches the one in the raw payload for the hash function.

Thanks again! I will merge #306 soon.

PR #299, opened 2018-05-18
Use Point or numerical type data to create a tuple for Nominatim view_box

### LINKED ISSUE #261: Nominatim: view_box parameter for Nominatim instance only works with string-represented numerical tuple
Hi and thanks for all your collective work on geopy. Really appreciate it and super helpful for data analytics! 

(Note, as an aside, that it's a little weird that the tuple is (longitude1, latitude1, longitude2, latitude2) instead of leading each 2-element pair within the 4-element tuple with latitude, but that seems to be [Nominatim's convention](https://github.com/openstreetmap/Nominatim/blob/81a7ea36db7d3b055acf25a36d2ea2c27af74765/test/bdd/api/search/params.feature), so it makes sense to keep it that way for the Class's parameters in geopy too - so that's not what this issue is about.)

Anyhow, the view_box parameter is a 4-element tuple of positive or negative integer or floating point data defining opposite corners of a rectangle in a 4-element python tuple(). If you use numerical data to create the tuple, you receive an error:

Setup and Call: 
```
box_tuple = tuple([180, 22, -180, -22])
geolocator = Nominatim(view_box=box_tuple)
location2 = geolocator.geocode("Peru")
```
 
Error:
```
~/anaconda3/lib/python3.6/site-packages/geopy/geocoders/osm.py in geocode(self, query, exactly_one, timeout, addressdetails, language, geometry)
    161         # `viewbox` apparently replaces `view_box`
    162         if self.view_box:
--> 163             params['viewbox'] = ','.join(self.view_box)
    164 
    165         if self.country_bias:

TypeError: sequence item 0: expected str instance, int found
```

If you use string data to create the tuple, no error, and the call to geocode() works as expected, resulting in a location returned that lies within the defined boundaries.

Setup and Call: 
```
box_tuple = tuple(["180", "22", "-180", "-22"])
geolocator = Nominatim(view_box=box_tuple)
location2 = geolocator.geocode("Peru")
location2
```

Output:
`Location(Perú, (-6.8699697, -75.0458515, 0.0))`

FWIW, my suggested fix would be to convert a numerical tuple to strings somehow, but I am not sure what the most efficient implementation or input handling would be, which is why I'm submitting this as an issue instead of coding an implementation.

I'm using geopy 1.11.0 with a Python 3.6.2 in a jupyter notebook.

Please let me know if I've been unclear or you need more info or have other questions. Thanks again, so much for all your work on this.

svalee:
this commit fixes #261

### DIFF
--- a/test/geocoders/nominatim.py
+++ b/test/geocoders/nominatim.py
@@ -215,16 +215,15 @@ class NominatimTestCase(GeocoderTestBase):
         )
 
     def test_view_box(self):
-        without_viewbox = self.geocode_run(
+        self.geocode_run(
             {"query": "Maple Street"},
             {"latitude": 36.809551, "longitude": -97.050604},
         )
-        for view_box in [(-0.9, 52, -0.11, 50),
-                         [Point(52, -0.9), Point(50, -0.11)],
-                         ("-0.9", "52", "-0.11", "50")]:
+        for view_box in [(-0.11, 52, -0.15, 50),
+                         [Point(52, -0.11), Point(50, -0.15)],
+                         ("-0.11", "52", "-0.15", "50")]:
             self.geocoder = Nominatim(view_box=view_box)
-            with_viewbox = self.geocode_run(
+            self.geocode_run(
                 {"query": "Maple Street"},
                 {"latitude": 51.5223513, "longitude": -0.1382104}
             )
-            self.assertNotEqual(with_viewbox, without_viewbox)
--- a/test/geocoders/nominatim.py
+++ b/test/geocoders/nominatim.py
@@ -207,3 +207,20 @@ class NominatimTestCase(GeocoderTestBase):
             'address',
             res
         )
+
+    def test_view_box(self):
+        box_tuple = tuple([180, 22, -180, -22])
+        geolocator = Nominatim(view_box=box_tuple)
+        location2 = geolocator.geocode("Peru")
+        self.assertEqual((location2.latitude, location2.longitude),
+                         (-6.8699697, -75.0458515))
+        box_tuple = [Point(22, 180), Point(-22, -180)]
+        geolocator = Nominatim(view_box=box_tuple)
+        location2 = geolocator.geocode("Peru")
+        self.assertEqual((location2.latitude, location2.longitude),
+                         (-6.8699697, -75.0458515))
+        box_tuple = tuple(["180", "22", "-180", "-22"])
+        geolocator = Nominatim(view_box=box_tuple)
+        location2 = geolocator.geocode("Peru")
+        self.assertEqual((location2.latitude, location2.longitude),
+                         (-6.8699697, -75.0458515))
--- a/test/geocoders/nominatim.py
+++ b/test/geocoders/nominatim.py
@@ -15,6 +15,12 @@ class NominatimTestCase(GeocoderTestBase):
     def setUpClass(cls):
         cls.geocoder = Nominatim()
 
+    def tearDown(self):
+        try:
+            del self.geocoder
+        except AttributeError:
+            pass
+
     def test_geocode(self):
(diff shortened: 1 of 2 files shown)


test/geocoders/nominatim.py
```diff
@@ -207,3 +207,20 @@ def test_missing_reverse_details(self):
             'address',
             res
         )
+
+    def test_view_box(self):
+        box_tuple = tuple([180, 22, -180, -22])
```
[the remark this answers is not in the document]
KostyaEsmukov: Another suggestion would be to turn this to a for loop, instead of copy-pasting the same code but with different formats of the view_box. 

Also it's more preferable to patch `self.geocoder` and use `self.geocode_run` instead of asserting the result manually. As `geocoder` is a class attribute, and modifying `self.geocoder` would change it on an instance rather than the class, you can simply assign it in the test and add a tearDown method which would delete the instance value to use the class's one again:
```
def tearDown(self):
    try:
        del self.geocoder
    except AttributeError:
        pass
```

A complete example:
```
    self.geocode_run(...)  # Show that without view_box the result is different.
    for view_box in [(180, 22, -180, -22),
                     [Point(22, 180), Point(-22, -180)],
                     ...]:
        self.geocoder = Nominatim(view_box=view_box)
        self.geocode_run(...)
```

I'm not sure if this is the best way, but for now it looks reasonable to me. What do you think?

test/geocoders/nominatim.py
```diff
@@ -209,18 +215,16 @@ def test_missing_reverse_details(self):
...
+        )
+        for view_box in [(-0.9, 52, -0.11, 50),
+                         [Point(52, -0.9), Point(50, -0.11)],
+                         ("-0.9", "52", "-0.11", "50")]:
+            self.geocoder = Nominatim(view_box=view_box)
+            with_viewbox = self.geocode_run(
+                {"query": "Maple Street"},
+                {"latitude": 51.5223513, "longitude": -0.1382104}
+            )
+            self.assertNotEqual(with_viewbox, without_viewbox)
```
svalee: This line might be redundant, should I remove it?

svalee:
Also, thanks for review, I was very anxious about the code and I'm very happy, that you said that it looks good!

test/geocoders/nominatim.py
KostyaEsmukov: Yes, I agree, this assertion looks redundant. Coordinates comparison is enough.

test/geocoders/nominatim.py
```diff
@@ -209,18 +215,16 @@ def test_missing_reverse_details(self):
...
-        box_tuple = tuple(["180", "22", "-180", "-22"])
-        geolocator = Nominatim(view_box=box_tuple)
-        location2 = geolocator.geocode("Peru")
-        self.assertEqual((location2.latitude, location2.longitude),
-                         (-6.8699697, -75.0458515))
+        without_viewbox = self.geocode_run(
+            {"query": "Maple Street"},
+            {"latitude": 36.809551, "longitude": -97.050604},
+        )
+        for view_box in [(-0.9, 52, -0.11, 50),
```
KostyaEsmukov: Maybe it's worth expanding the longitudes a little? The result `-0.138` doesn't lie within the `-0.11...-0.9` range. I think something like `-0.15..-0.11` is less likely to accidentally break.

KostyaEsmukov:
> Also, thanks for review, I was very anxious about the code and I'm very happy, that you said that it looks good!

No need for anxiety, you're doing great! Thanks for your efforts, I appreciate that a lot!

svalee:
> No need for anxiety, you're doing great! Thanks for your efforts, I appreciate that a lot!
Thanks a lot! I really appreciate your positive attitude.

PR #300, opened 2018-06-04
Proxies can be sent as a string

### LINKED ISSUE #290: Proxies: accept as string
I think that accepting a single string for the `proxies` param would be great. The urllib2-style dict objects with schemes might be confusing to the users of geocoders.

    # Current usage:
    geolocator = Nominatim(scheme='https', proxies={'https': 'http://192.0.2.1:8080'})
    # Proposed usage:
    geolocator = Nominatim(proxies='http://192.0.2.1:8080')

~~#288 should be merged first.~~

svalee:
this commit fixes #290

### DIFF
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -188,8 +188,8 @@ class Geocoder(object):
 
         if self.proxies:
             if isinstance(self.proxies, str):
-                self.scheme = self.proxies.split('://')[0]
-                self.proxies = {self.scheme: self.proxies}
+                self.proxies = {self.scheme if self.scheme else
+                                self.proxies.split('://')[0]: self.proxies}
 
             opener = build_opener_with_context(
                 self.ssl_context,
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -188,6 +188,7 @@ class Geocoder(object):
 
         if self.proxies:
             if isinstance(self.proxies, str):
+                print(self.scheme, 'scheme')
                 self.proxies = {self.scheme if self.scheme else
                                 self.proxies.split('://')[0]: self.proxies}
 
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -188,9 +188,7 @@ class Geocoder(object):
 
         if self.proxies:
             if isinstance(self.proxies, str):
-                print(self.scheme, 'scheme')
-                self.proxies = {self.scheme if self.scheme else
-                                self.proxies.split('://')[0]: self.proxies}
+                self.proxies = {self.scheme: self.proxies}
 
             opener = build_opener_with_context(
                 self.ssl_context,
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -187,7 +187,7 @@ class Geocoder(object):
                             else options.default_ssl_context)
 
         if self.proxies:
-            if isinstance(self.proxies, str):
+            if isinstance(self.proxies, string_compare):
                 self.proxies = {self.scheme: self.proxies}
 
             opener = build_opener_with_context(
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -188,7 +188,7 @@ class Geocoder(object):
 
         if self.proxies:
             if isinstance(self.proxies, string_compare):
-                self.proxies = {self.scheme: self.proxies}
+                self.proxies = {'http': self.proxies, 'https': self.proxies}
 
             opener = build_opener_with_context(
                 self.ssl_context,
--- a/geopy/geocoders/base.py
+++ b/geopy/geocoders/base.py
@@ -70,6 +70,8 @@ class options(object):
             If specified, tunnel requests through the specified proxy.
             E.g., ``{"https": "192.0.2.0"}``. For more information, see
             documentation on :class:`urllib.request.ProxyHandler`.
+            Can be sent as a string ``"https://192.0.2.0"``, this will
+            automatically set the scheme to the ``https``
 
         default_scheme
             Use ``'https'`` or ``'http'`` as the API URL's scheme.
@@ -185,6 +187,10 @@ class Geocoder(object):
                             else options.default_ssl_context)
 
         if self.proxies:
+            if isinstance(self.proxies, str):
+                self.scheme = self.proxies.split('://')[0]
+                self.proxies = {self.scheme: self.proxies}
+
             opener = build_opener_with_context(
                 self.ssl_context,
                 ProxyHandler(self.proxies),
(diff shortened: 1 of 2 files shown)


geopy/geocoders/base.py
```diff
@@ -185,6 +187,10 @@ def __init__(
                             else options.default_ssl_context)
 
         if self.proxies:
+            if isinstance(self.proxies, str):
+                self.scheme = self.proxies.split('://')[0]
```
KostyaEsmukov: First of all, `proxies` shouldn't affect `self.scheme` in any way.

Proxy scheme and URL scheme are totally different and doesn't need to match.

For example, for `proxies = 'http://my_proxy_server:8080'` and `self.scheme = 'https'` the client would send a [CONNECT][1] method to the proxy (in plaintext), and then start TLS negotiation with the remote using that TCP connection.

Just remove this line and it will be fine.

[1]: https://tools.ietf.org/html/rfc7231#section-4.3.6

geopy/geocoders/base.py
```diff
@@ -185,6 +187,10 @@ def __init__(
...
+                self.proxies = {self.scheme: self.proxies}
```
KostyaEsmukov: Given that we verify above that `self.scheme` is either `http` or `https`, I would suggest to specify proxy for both schemes instead of the current one.

Consider this situation:
```
g = SomeGeocoder(proxies='http://my_proxy_server:8080', scheme='https')
g.scheme = 'http'
```

After assignment proxy wouldn't be used, which is unexpected.

geopy/geocoders/base.py
```diff
@@ -185,6 +187,10 @@ def __init__(
                             else options.default_ssl_context)
 
         if self.proxies:
+            if isinstance(self.proxies, str):
```
KostyaEsmukov: For compatibility with py2, `geopy.compat.string_compare` should be used instead of `str`.

geopy/geocoders/base.py
```diff
@@ -70,6 +70,8 @@ class options(object):
             If specified, tunnel requests through the specified proxy.
             E.g., ``{"https": "192.0.2.0"}``. For more information, see
             documentation on :class:`urllib.request.ProxyHandler`.
+            Can be sent as a string ``"https://192.0.2.0"``, this will
+            automatically set the scheme to the ``https``
```
KostyaEsmukov: To be honest, I don't like it 🤔. But I'm not ready to propose a better phrasing for now as well.

Let's just keep it, but I'll get to this later after merging.

svalee:
> Thank you, I think this change would make API much better and more pleasant to use.

Thanks as usual for review and friendliness. Yeah, it's definitely going to be easier to use.

> Just a couple of changes and it should be ready to go. As for the doc – I suggest you to not care about it for now. I wanted to rewrite the proxy section anyway, and this would be another good reason to finally get to it.

Thank you very much, putting the right thought into the docs is still the most tough part for me.

> PS. Points for a good test! grinning

Thanks, I didn't do much just copied the exsiting test.

It took me some time to understand that there is a default scheme. This is why I put 2 tests, one with the https which is default, and one with http. But they both set explicitly, I think we can omit setting it explicitly in the https test, should we?

geopy/geocoders/base.py
KostyaEsmukov: I guess I wasn't clear enough on this one.

This is what I meant:
```
self.proxies = {'http': self.proxies, 'https': self.proxies}
```

That way my snippet above will work correctly.

test/test_proxy.py
```diff
@@ -77,14 +77,28 @@ def test_ssl_context_with_proxy_is_respected(self):
         self.assertIn('SSL', str(cm.exception))
         self.assertEqual(1, len(self.proxy_server.requests))
 
-    def test_geocoder_constructor_uses_str_proxy(self):
+    def test_geocoder_constructor_uses_str_http_proxy(self):
         base_http = urlopen(self.remote_website_http, timeout=self.timeout)
         base_html = base_http.read()
-        geocoder_dummy = DummyGeocoder(proxies=self.proxy_url,
+        geocoder_dummy = DummyGeocoder(proxies=self.proxy_url, scheme='http',
```
KostyaEsmukov: No, I don't think we need to test two schemes in two almost identical integration tests, which aren't cheap with regards to time. 

It would be better to keep the previous single test, and add a new *unit* test here. Something like this:
```
g = DummyGeocoder(proxies=self.proxy_url, scheme='http')
self.assertDictEqual(g.proxies, {'http': self.proxy_url, 'https': self.proxy_url})
```

That should be enough. We already tested that the `g.proxies` dict is respected in the integration tests, but we didn't check if both schemes are respected – which is exactly what this test does.

svalee: I am not sure if I get this right, but this will assert error:

`self.assertDictEqual(g.proxies, {'http': self.proxy_url, 'https': self.proxy_url})`

So I added 2 DummyGeocoders to check if both schemes are respected.

KostyaEsmukov: Have you by chance missed this comment? https://github.com/geopy/geopy/pull/300#discussion_r193676957

KostyaEsmukov: Also it would've been better if it was in a separate `test_*` function.

svalee: Oh yeah, I did miss this comment, pardon my inattention.

test/test_proxy.py
```diff
@@ -76,3 +76,21 @@ def test_ssl_context_with_proxy_is_respected(self):
             geocoder_dummy.geocode(self.remote_website_https)
         self.assertIn('SSL', str(cm.exception))
         self.assertEqual(1, len(self.proxy_server.requests))
+
+    def test_geocoder_constructor_uses_str_proxy(self):
+        base_http = urlopen(self.remote_website_http, timeout=self.timeout)
+        base_html = base_http.read()
+        geocoder_dummy = DummyGeocoder(proxies=self.proxy_url, scheme='http',
```
KostyaEsmukov: Is there any reason to set the `scheme` kwarg explicitly? If not, could you please remove that?

test/test_proxy.py
```diff
@@ -76,3 +76,21 @@ def test_ssl_context_with_proxy_is_respected(self):
...
+                                       timeout=self.timeout)
+        print(geocoder_dummy.scheme, geocoder_dummy.proxies, 'proxy')
```
KostyaEsmukov: 🙃

svalee: whoops 😅

KostyaEsmukov:
Thanks! 🎉

svalee:
Yay! Thank you very much, I'm looking forward to continue my work with you (can we call it this way?) 🎂

KostyaEsmukov:
Of course, you're very welcome! I'm very grateful to you for your efforts on making geopy better. ✨🦄

PR #303, opened 2018-06-09
Update What3Words to v2 API

KostyaEsmukov:
Supersedes #226.

This updates What3Words geocoder to API v2. The following might affect the users of this geocoder class:
- `Location.raw` results are different.
- `http` scheme is now disallowed.

The rest should stay the same.

### DIFF
--- a/geopy/geocoders/what3words.py
+++ b/geopy/geocoders/what3words.py
@@ -17,12 +17,11 @@ class What3Words(Geocoder):
     """What3Words geocoder.
 
     Documentation at:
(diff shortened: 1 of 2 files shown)


karussell:
If you care for open data and open source you should not support them. Read more about it [here](https://blog.ldodds.com/2016/06/14/what-3-words-jog-on-mate/) and [here](https://wiki.openstreetmap.org/w/index.php?title=What3words&oldid=1421526). (You could use plus code as alternative which is [open](https://github.com/google/open-location-code)).

PR #304, opened 2018-06-12
Add HERE geocoder

deeplook:
This is a wrapper around [HERE's geocoder API](https://developer.here.com/documentation/geocoder/) (incl. reverse geocoding) which can also be used to retrieve various kinds of polygons (one test is included). Additional tests might follow later.


geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+        # FIXME(?): add multi-reverse geocode
+        # FIXME(?): add landmark geocode
+
+    def geocode(
+            self,
+            query,
+            exactly_one=True,
+            language=None,
+            additional_data=False,
+            other=None,
```
KostyaEsmukov: Could you please remove this kwarg? As stated in [CONTRIBUTING.md][1], allowing to pass arbitrary parameters is discouraged, as it's prone to break users' code on API version upgrades.

[1]: https://github.com/geopy/geopy/blob/602150a3c9c4ec63206325db4e50a3c9b40b43ac/CONTRIBUTING.md#improving-a-geocoder

geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+        :param str query: The address or query you wish to geocode.
+
+            For a structured query, provide a dictionary whose keys
+            are one of: `city`, `county`, `district`, `country`, `state`,
+            `street`, `housenumber`, or `postalcode`.
+
+        :param bool exactly_one: Return one result or a list of results, if
+            available.
+
+        :param int timeout: Time, in seconds, to wait for the geocoding service
```
KostyaEsmukov: This is out of order. `timeout` kwarg is the last one, this doc should be too.

geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+            language=None,
+            additional_data=False,
+            other=None,
+            timeout=DEFAULT_SENTINEL
+    ):
+        """
+        Return a location point by address.
+
+        Documentation at:
+            https://developer.here.com/documentation/geocoder/topics/resource-geocode.html
```
KostyaEsmukov: `geocode`/`reverse` methods usually don't contain links to the relevant API docs, given that there's already a link in the class's docstring. Also the links tend to change, so having a single link is easier for maintenance and is less cluttering.

What do you think? I guess it was helpful to have these links around with the `other` kwarg, but without it they seem to be less useful.

geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+        :rtype: ``None``, :class:`geopy.location.Location` or a list of them, if
+            ``exactly_one=False``.
+        """
+        point = self._coerce_point_to_string(query)
+        params = {
+            'app_id': self.app_id,
+            'app_code': self.app_code,
+            'maxresults': 1,
+            'mode': 'retrieveAddresses',  # can be overwritten using `other`
+            'prox': point + ',1000',
```
KostyaEsmukov: Is radius value required? If not, it would be better to not specify it explicitly. But in any way it should be a separate kwarg, so it could be changed.

geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+            elif status_code == 503:
+                raise GeocoderUnavailable(err)
+            else:
+                raise GeocoderServiceError(err)
+
+        try:
+            resources = doc['Response']['View'][0]['Result']
+        except IndexError:
+            resources = None
+        if resources is None or not len(resources):
```
KostyaEsmukov: I think this is equal to `if not resources:` in this case.

geopy/geocoders/here.py
```diff
@@ -0,0 +1,281 @@
...
+            country = addr.get('Country', '').strip(stripchars)
+
+            city_state = join_filter(", ", [city, state])
+            place = join_filter(" ", [city_state, zipcode])
+            location = join_filter(", ", [address, place, country])
+
+            display_pos = resource['Location']['DisplayPosition']
+            latitude = display_pos['Latitude'] or None
+            longitude = display_pos['Longitude'] or None
+            if latitude and longitude:
```
KostyaEsmukov: Does the documentation say that these might be missing? Sorry I didn't look it myself.

If it doesn't, maybe we could assume that it's always there and fail when it's not? For example, `float(None)` would raise a TypeError exception. So instead we can simply do this:
```
latitude = float(display_pos['Latitude'])
longitude = float(display_pos['Longitude'])
```

Also the `or None` part would eat up `0.0`.

test/geocoders/here.py
```diff
@@ -0,0 +1,138 @@
...
+            {"query": u("\u30d1\u30ea")},
+            {"latitude": 48.85718, "longitude": 2.34141}
+        )
+
+    def test_geocode_shapes(self):
+        """
+        Here.geocode using additional data parameter (postal code shapes)
+        """
+        address_string = "435 north michigan ave, chicago il 60611 usa"
+        res = self._make_request(
```
KostyaEsmukov: Could you please remove `_make_request` usages?

```
res = self.geocode_run(
    {"query": address_string, "additional_data": "IncludeShapeLevel,postalCode"},
    {"latitude": 41.89035, "longitude": -87.62333},
)
shape_value = res.raw['Location']['Shape']['Value']
self.assertTrue(shape_value.startswith('MULTIPOLYGON ((('))
```

deeplook:
Thanks for giving very detailed comments so quickly! I'll address them in-place, but in general, 10-15 days should be more than enough time! :)

geopy/geocoders/here.py
deeplook: I totally understand the motivation. From all the existing params I've taken only `language` as the one available for most/all geocoders (I guess) and `additional_data` as this allows for getting shapes, too, which I think is rather unusual. The problem is: there are *many* more params, and only `additional_data` can have [17 different keys](https://developer.here.com/documentation/geocoder/topics/resource-params-additional.html). For the geocode API there are [28 more](https://developer.here.com/documentation/geocoder/topics/resource-geocode.html), plus [21 for the reverse geocode API](https://developer.here.com/documentation/geocoder/topics/resource-reverse-geocode.html)! And I haven't even added the multi-reverse geocode and landmark geocode APIs... (see FIXMEs).

So I felt like having an `other` parameter is the most efficient way of getting something out quickly, being well aware that this can not be documented nicely inside geopy, but only on the external REST API documentation. In this case I was told this is unlikely to change since there will be a new generation of these HERE APIs under a different label, anyway, but I'll double-check, again. So, if possible I'd really avoid putting in such a huge amount of work, but instead try to add a few more interesting test cases. Would that be an option? How is that handled for other rich APIs?

geopy/geocoders/here.py
deeplook: Ok, sure.

geopy/geocoders/here.py
deeplook: Sounds right. Will change, if so.

geopy/geocoders/here.py
deeplook: Ok, will test and make a separate param if possible.

test/geocoders/here.py
deeplook: Ok, sure. BTW, I wondered if it would be meaningful to add tests on the resulting shapes, but that would be easier with additional packages (like `shapely.wkt` or something smaller). What's the policy of adding other packages for testing only?

geopy/geocoders/here.py
KostyaEsmukov: A lot of geocoding services provide a rich set of parameters, options and API methods. But more than 28 parameters indeed look rather unusual though.

Generally implementing full support for all parameters and methods is not a goal in geopy. Implementing basic forward/reverse geocoding is usually enough for a lot of use-cases. For missing features [the doc suggests][semver] to open PRs for adding those (yup, the wording might have been more explicit with regards to missing API parameters).

While it's possible to explicitly define all of these peculiar parameters, a really small subset of users would really need them. And even then they would probably be better off using a specialized SDK package for that geocoder, which goal is exactly to match the API as close as possible. But, having all these rarely needed parameters implemented in geopy, would force us to rename the geocoder class once they change their API version (like it was done with `Google` and `GoogleV3` a while ago), if they drop or change some of the parameters; in order to not silently break the code of people who did use these parameters. But the large group of people, who didn't, would have to rename the class to upgrade, while other changes from them are not needed.

So the general approach is trying to keep the geocoder's interface sufficient enough for typical use-cases. Easier API version upgrades is more of a consequence of that rather than a goal.

The exact set of parameters is up to you. For me what's already there looks fine.

[semver]: https://geopy.readthedocs.io/en/1.14.0/#semver

geopy/geocoders/here.py
KostyaEsmukov: It's a matter of taste, indeed. I'm fine with keeping them if you think they're helpful there.

test/geocoders/here.py
KostyaEsmukov: I'm not sure that it's needed. Geopy doesn't work with shapes, so it's the API's responsibility to ensure that the returned shapes are valid. For geopy it should be enough to assume that if it *looks* like a shape (like you did with checking the prefix), then it's a valid shape.

As for the policy, the less deps the better :)

geopy/geocoders/here.py
deeplook: Well, I can look up some more that might be useful more generally. The point is that allowing only a specific subset makes it *impossible* to use the rest, while using a more generic approach only makes it *harder* to document. I would favour the first giving people more flexibility, but it's your package. :)

test/geocoders/here.py
deeplook: Sounds good.

geopy/geocoders/here.py
KostyaEsmukov: Not impossible exactly. I can come up with at least one dirty workaround involving `threading.local()` and overriding of `_call_geocoder`. This is *very* ugly, but possible. :)

Please don't get me wrong: I would love to give that flexibility. But not at a cost of painful upgrades for the most of the users. It's definitely not just a matter of documentation. It's a matter of losing confidence over the scope of provided geocoder capabilities, which is crucial for preserving backwards compatibility when upgrading underlying APIs.

I've had this issue in mind for a long time, but now as you brought it, I've just created #305 to track this. I hope eventually we will come up with a decent solution to this.

deeplook:
@KostyaEsmukov Because you asked me for changed within 10-15 days: I had kind of no time to work on this, until now, but I'll give it a try until/including the weekend to come.

KostyaEsmukov:
Thank you! I'm off my schedule as well: hopefully I'll deal with the rest of the 1.15 blockers by the middle of next week.

deeplook:
Making some progress with updates to my PR, I wonder if it is such a good idea to keep geopy's `exactly_one` parameter *and* HERE's `maxresults` one? The former is part of the official plugin interface, I guess, but the latter is more general. I can have both, but it needs some choice of precedence and explanation.

KostyaEsmukov:
> I wonder if it is such a good idea to keep geopy's `exactly_one` parameter *and* HERE's `maxresults` one?

Yep, both should be present. See this comment for rationale and precedence explanation: https://github.com/geopy/geopy/pull/281#discussion_r180079906

> In fact, for `maxresults` to make most sense I would need to have a `pageinformation` param for implementing some simple paging mechanism, too. But I can't find anything on "paging" in this package. Has really no other plugin implemented paging?

I'd suggest to not bother about pagination. Looping through result pages is very rarely (if ever) needed for geocoding results. Usually blindly selecting the first result (i.e. with `exactly_one=True` by default) is good enough.

`maxresults` is perfectly viable even without `pageinformation`. For example, to limit the default page size (what is it, ~20?) to 5, which might be reasonable for live-search suggestions.

KostyaEsmukov:
Hey @deeplook , how is it going? I'd like to cut a release this weekend. Is there any chance you could push the changes soon? If it's inconvenient or stressful for you, we can simply move this feature to the next version. Thanks anyway!

deeplook:
I can push what I have today, but it won't include a few more complicated tests I'd like to provide, too. From my point of view that should be still ok, and I would rather see the code released now and provide these tests, later. What do you think?

deeplook:
@KostyaEsmukov I've pushed what I have. Looks like travis isn't happy yet (env. vars?), but I can't do much more now, as I need to pack for a vacation. If you can finalize this, that would be awesome...

KostyaEsmukov:
Don’t worry, I’ll do that for you. Thanks and have a great vacation!

KostyaEsmukov:
The aforementioned nitpicking commit: 3d689bc937952714833427d9c96c2dbe015e60b2

PR #316, opened 2018-07-27
Nominatim: show a warning when default user_agent is used

### LINKED ISSUE #314: Nominatim geopy.exc.GeocoderInsufficientPrivileges: HTTP Error 403: Forbidden with default user-agent
When using default user-agent: `geopy/x.y` the Nominatim service returns 403: `geopy.exc.GeocoderInsufficientPrivileges: HTTP Error 403: Forbidden`. Setting a custom user-agent fixes the issue.

Seems someone is bombing the osm with geopy. What about to make the `user_agent` argument compulsory, so everyone has to make own `user-agent`? Or it should be mentioned in the docs to set it up always, at least.

See also #262

KostyaEsmukov:
[Nominatim's Usage Policy](https://operations.osmfoundation.org/policies/nominatim/) says that each application must provide their own user_agent. geopy allowed to use Nominatim with stock geopy's user_agent, but lately people started to report receiving 403 HTTP errors (see #314), probably caused by someone bombing Nominatim with requests using the stock geopy user-agent.

This PR adds a warning when stock user-agent is used against `nominatim.openstreetmap.org`. In geopy 2.0 this will be an exception instead of a warning. Also all examples have been updated to include the custom user_agent.

cc @xmedeko 

Closes #314

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ To geolocate a query to an address and coordinates:
 .. code:: python
 
     >>> from geopy.geocoders import Nominatim
-    >>> geolocator = Nominatim()
+    >>> geolocator = Nominatim(user_agent="specify_your_app_name_here")
     >>> location = geolocator.geocode("175 5th Avenue NYC")
     >>> print(location.address)
     Flatiron Building, 175, 5th Avenue, Flatiron, New York, NYC, New York, ...
@@ -88,7 +88,7 @@ To find the address corresponding to a set of coordinates:
 .. code:: python
 
     >>> from geopy.geocoders import Nominatim
-    >>> geolocator = Nominatim()
+    >>> geolocator = Nominatim(user_agent="specify_your_app_name_here")
     >>> location = geolocator.reverse("52.509669, 13.376294")
     >>> print(location.address)
     Potsdamer Platz, Mitte, Berlin, 10117, Deutschland, European Union
--- a/geopy/distance.py
+++ b/geopy/distance.py
@@ -79,7 +79,7 @@ calculate the length of a path::
 
     >>> from geopy import Nominatim
     >>> d = distance.distance
-    >>> g = Nominatim()
+    >>> g = Nominatim(user_agent="specify_your_app_name_here")
     >>> _, wa = g.geocode('Washington, DC')
     >>> _, pa = g.geocode('Palo Alto, CA')
     >>> print((d(ne, cl) + d(cl, wa) + d(wa, pa)).miles)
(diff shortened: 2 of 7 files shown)


xmedeko:
Very well made, thanks.

rpahut:
I am not convinced this change, and the ones following out of it, are in line with the nominatim policy.
First off, it is entirely hypothetical that #314 was _caused_ by user-agent. It might have as well been an individual block xmedeco earned for himself, and then was able to bypass by changing the user-agent. On the other hand, if there is a flaw that really earned the entire version of geopy a block, that would be obscured by this change, and it may remain unfixed to this day.

The reason API requires the user-agent is to be able to identify the agent it is talking to, which in this case is geopy. We as users have little control over the actual exchange.
It might seem beneficial to bypass the user-agent concern altogether, but since it dismantles the regulation mechanisms of the service we use here, it is not actually in interests of anyone intending to use the library responsibly. I mean for sure, it shouldn't force the user to set the user-agent string to arbitrary, meaningless values, as it currently does.

KostyaEsmukov:
> First off, it is entirely hypothetical that #314 was _caused_ by user-agent. 

It's not. Back then, there was a clear, 100% correlation with errors and a stock user-agent being used.

> The reason API requires the user-agent is to be able to identify the agent it is talking to

Correct

> which in this case is geopy. 

Incorrect. It's the application/script that calls geopy, which should properly announce its name to Nominatim per their tos.

Which is why geopy forces the user to supply a proper user-agent identifying their application.

rpahut:
Sorry bro just calling a thing "incorrect", in TOS or otherwise, doesn't make it so. If your goal is to build an internally consistent system the decisions must be built upon a solid verifiable logic, not arbitrary statements. The logic of the situation here is that the application/script uses geopy, and geopy communicates with the server on its behalf. The situation makes the app an user, and geopy _is its agent_.

Then, consider what the original issue was. If an app using geopy _did_ manage to strain the server to the point where it had to start blocking - that is he issue that asks to be addressed. Probably _between the library and the server_ where it happens. This PR instead deals with the _symptom_, pushes the responsibiliy off to the user, and destroys the meaning of the user-agent string. Note that this might have restored the connection, but it did so just by side stepping the servers ability to identify the misbehaving agent and cut it off.

> there was a clear, 100% correlation

That is not how I see it being described, anywhere. In this PR description, you used the word "probably".
Then, is this still an issue, or was at any point since it was originally reported?
Then still, who is to say that geopy didn't actually earn that block by misbehaving?

PR #323, opened 2018-08-31
add mapbox geocoder

williamhammond:


geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ValueError("Invalid point")
+
+    def geocode(
+            self,
+            query,
+            exactly_one=True,
+            proximity=None,
+            country=None,
+            bbox=None,
+            timeout=DEFAULT_SENTINEL
```
KostyaEsmukov: `timeout` should be the third positional argument (i.e. right after `exactly_one`), to match the signature of the base method: 

https://github.com/geopy/geopy/blob/0a3a60bd6f9cb570e346ffef7953babf630ae8a3/geopy/geocoders/base.py#L368

Also please add trailing commas everywhere for cleaner diffs in future.

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,209 @@
...
+
+        :param bbox: The bounding box of the viewport within which
+            to bias geocode results more prominently.
+        :type bbox: list or tuple
+
+        :param int timeout: Time, in seconds, to wait for the geocoding service
+            to respond before raising a :class:`geopy.exc.GeocoderTimedOut`
+            exception. Set this only if you wish to override, on this call
+            only, the value set during the geocoder's initialization.
+        """
```
KostyaEsmukov: `rtype` is missing: https://github.com/geopy/geopy/blob/0a3a60bd6f9cb570e346ffef7953babf630ae8a3/geopy/geocoders/tomtom.py#L102

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,209 @@
...
+        params = {}
+
+        params['access_token'] = self.api_key
+
+        if query is None or query == '':
+            raise GeocoderQueryError('query must be set')
```
KostyaEsmukov: I believe this check is unnecessary. Let the API return an error in this case (if this is even an error). It's not very helpful, yet it makes code harder to read.

Also the `query` must be formatted using `self.format_string`. See for example: 

https://github.com/geopy/geopy/blob/0a3a60bd6f9cb570e346ffef7953babf630ae8a3/geopy/geocoders/tomtom.py#L105

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,209 @@
...
+                raise GeocoderQueryError(
+                    "bbox must be a four-item iterable of lat,lon,lat,lon"
+                )
+            params['bbox'] = self._format_bbox_param(bbox)
+
+        if country:
+            if len(country) != 2:
+                raise GeocoderQueryError(
+                    "country must be specified as ISO 3166-1 alpha-2 country code."
+                )
```
KostyaEsmukov: I would remove this check too. We don't really care about the length, and we cannot know for sure the supported languages anyway, to make the proper validation. So I'd let the backed validate this value.

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,209 @@
...
+            params['country'] = country
+
+        if proximity:
+            if len(proximity) != 2:
+                raise GeocoderQueryError(
+                    "proximity must be a two-item iterable of lat,lon"
+                )
+            params['proximity'] = self._format_proximity_param(proximity)
+
+        request_url = self.api + "/" + quote(query) + ".json/"
```
KostyaEsmukov: I believe it should be `quote(query.encode('utf-8'))`.

This concatenation might look better (IMO) if we do the following:
1. Introduce class attribute: `geocode_path = '/geocoding/v5/mapbox.places/%(query)s.json/'`
2. In `__init__` do: `self.api = "%s://%s%s" % (self.scheme, self.domain, self.geocode_path)`
3. Here do the following:

        quoted_query = quote(query.encode('utf-8'))
        url = "?".join((self.api % dict(query=quoted_query),
                        urlencode(params)))

You may find an example of this approach here: https://github.com/geopy/geopy/blob/0a3a60bd6f9cb570e346ffef7953babf630ae8a3/geopy/geocoders/tomtom.py#L117

It's just a suggestion. This concatenation might be fine as it is (except the missing `.encode('utf-8')`).

williamhammond:
Thanks for the feedback! I think i got everything you pointed out, let me know if i missed anything.

README.rst
```diff
@@ -46,6 +46,7 @@ The various geocoder classes are located in `geopy.geocoders`_.
 .. _SmartyStreets: https://smartystreets.com/products/liveaddress-api
 .. _GeocodeFarm: https://www.geocodefarm.com/
 .. _Here: https://developer.here.com/documentation/geocoder/
+.. _MapBox: https://www.mapbox.com/api-documentation/#geocoding
```
KostyaEsmukov: In RST this is just a named link, which should be referenced somewhere. More specifically, it must be added to the paragraph above this list.

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,215 @@
...
+            user_agent=None,
+            ssl_context=DEFAULT_SENTINEL,
+            domain='api.mapbox.com',
+    ):
+        """
+        :param str api_key: The API key required by Mapbox to perform
+            geocoding requests. API keys are managed through Mapox's account
+            page (https://www.mapbox.com/account/access-tokens).
+
+        :param str domain: base api domain for mapbox
```
KostyaEsmukov: The order of params in the docstring should match the order of params of the function.

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,215 @@
...
+
+        :param dict proxies:
+            See :attr:`geopy.geocoders.options.default_proxies`.
+
+        :param str user_agent:
+            See :attr:`geopy.geocoders.options.default_user_agent`.
+
+        :type ssl_context: :class:`ssl.SSLContext`
+            :param ssl_context:
+                See :attr:`geopy.geocoders.options.default_ssl_context`.
```
KostyaEsmukov: Indentation is off by 1 here and the line above. I suppose this would be an error when generating docs. 

These lines should be aligned like that:
https://github.com/geopy/geopy/blob/0a3a60bd6f9cb570e346ffef7953babf630ae8a3/geopy/geocoders/tomtom.py#L50-L52

geopy/geocoders/mapbox.py
```diff
@@ -0,0 +1,215 @@
...
+        :type proximity: list of tuple
+
+        :param string country: Country to filter result in form of
+            ISO 3166-1 alpha-2 country code.
+
+        :param bbox: The bounding box of the viewport within which
+            to bias geocode results more prominently.
+        :type bbox: list or tuple
+
+        :param int timeout: Time, in seconds, to wait for the geocoding service
```
KostyaEsmukov: The order doesn't match the parameters. `timeout` doc should go after `exactly_one`.

williamhammond:
Looks like the build failed non-mapbox tests, is there a way i can manually restart it?

KostyaEsmukov:
> Looks like the build failed non-mapbox tests, is there a way i can manually restart it?

None that I am aware of, except a git push.

I restarted the failed jobs manually and they passed.

Issue #326, opened 2018-10-07
support for Geonames reverse findNearby in addition to findNearbyPlaceName


biokomiker:
Geonames reverse supports two variants:

http://api.geonames.org/ **findNearby** ?lat=46.4964&lng=9.28403&username=demo

<geonames><geoname><toponymName>Pizzo Tambo</toponymName><name>Pizzo Tambo</name><lat>46.49703</lat><lng>9.2834</lng><geonameId>2658429</geonameId><countryCode>CH</countryCode><countryName>Switzerland</countryName><fcl>T</fcl><fcode>PK</fcode><distance>0.08547</distance></geoname></geonames>

http://api.geonames.org/ **findNearbyPlaceName** ?lat=46.4964&lng=9.28403&username=demo

<name>Montespluga</name>

The output differs the first one being more useful for reverse tagging pictures, the second one being implemented in geopy.

I suggest to add some code in order to be also able to make the findNearby searches via geopy. It just requires some minor changes in geonames.py and defaults to the old behaviour.

Greetings !

```
    reverse_path = '/findNearbyPlaceNameJSON'
    # add findNearby
    findNearby_path = '/findNearbyJSON'
```

        self.api_reverse = (
            "%s://%s%s" % (self.scheme, domain, self.reverse_path)
        )
        self.api_reverse_findNearby = (
            "%s://%s%s" % (self.scheme, domain, self.findNearby_path)         
        )  # added to allow search findNearby

    def reverse(
            self,
            query,
            exactly_one=DEFAULT_SENTINEL,
            timeout=DEFAULT_SENTINEL,
            find_nearby_type='findNearbyPlaceName',      # added to allow search findNearby
    ):

        if find_nearby_type=='findNearby':  # added to allow search findNearby
            url = "?".join((self.api_reverse_findNearby, urlencode(params)))
        else:
            url = "?".join((self.api_reverse, urlencode(params)))
        logger.debug("%s.reverse: %s", self.__class__.__name__, url)
        return self._parse_json(
            self._call_geocoder(url, timeout=timeout),
            exactly_one
        )
```
```

KostyaEsmukov:
Support for `findNearby` has been added in #327 and ae1d6c493.

PR #327, opened 2018-10-08
 added a timezone parameter to geonames

### LINKED ISSUE #163: add timezone to geonames
See:
http://www.geonames.org/export/web-services.html#timezone

svalee:
this fixes #163

Hey, it's been a while. I wasn't sure how to correctly implement this thing and I can see that there is a lot of work here to do. As far as an I can see, the timezone url is very different from geocode and reverse, however it has the same api as reverse, so I decided to put it there. I didn't write tests for this, for the same reason, I wanted to hear from you how are you seeing this timezone parameter?


KostyaEsmukov:
Hey, it's always good to see your PRs! I hope you're doing well.

Actually there's a related issue discussed in #326. I even think these two might be resolved at once.

As for the `/timezone` method, I don't think that it should be combined with the `reverse` method, because the returned data is semantically different. The plain `reverse` is intended to return a full address of the point, while the timezone is a different thing. In fact, GoogleV3 implements a similar functionality, so I think it could be used as an example here: https://github.com/geopy/geopy/blob/a4e663a16291e02975a83426bb0a5ede00c70a0e/geopy/geocoders/googlev3.py#L330

As for the `/findNearby` method, discussed in #326, well, I see at least two possible ways to implement it:
1. As a new single choice-like (or a boolean?) parameter for the `reverse` method,
2. As a new distinct method in the Geocoder class.

The first approach might look confusing (especially given that the two API methods, being called by the same single `reverse` function, have different sets of parameters), but it would allow to use the new API method with the other libraries/frameworks which assume that reverse geocoding can be done only one way (I suppose `geopandas` behaves this way).

The second approach might look cleaner from the geocoder's interface point of view, although it may still be confusing that a geocoder provides two distinct methods both doing reverse geocoding (*it is* confusing to me why geoname does actually provide so many different methods which could just be parameters for a single reverse method).

I haven't made up my mind yet on which way to go there, I would be glad to hear your thoughts on this one. I tend to incline towards the first option, yet I'm not certain about it. But anyway, this is out of the scope of the issue you're trying to resolve with this PR, so you may leave this out and work specifically on the timezone method instead.

svalee:
Hi again, sorry for the wait, I've been really busy lately.
I have a question for timezone. Is it okay to do it like this, because api supports some additional optional keys such as radius lang date, but if we will return pytz object they're useless.
For example here what it can return with a request like this:
`http://api.geonames.org/timezoneJSON?lat=47.01&lng=10.2&radius=10&lang=ru&date=2018-01-01&username=demo`


`{"sunrise":"2018-10-18 07:40","lng":10.2,"countryCode":"AT","gmtOffset":1,"rawOffset":1,"sunset":"2018-10-18 18:27","timezoneId":"Europe/Vienna","dstOffset":2,"dates":[{"date":"2018-01-01","sunrise":"2018-01-01 08:05","sunset":"2018-01-01 16:40"}],"countryName":"Австрия","time":"2018-10-18 18:12","lat":47.01}`

Same goes for findNearby, but I think it doesn't matter that much or does it?

geopy/geocoders/geonames.py
```diff
@@ -5,11 +5,17 @@
     ConfigurationError,
     GeocoderInsufficientPrivileges,
     GeocoderServiceError,
-)
+    GeocoderParseError)
```
KostyaEsmukov: The closing parent should be on its own line.

geopy/geocoders/geonames.py
```diff
@@ -87,6 +95,12 @@ def __init__(
         self.api_reverse = (
             "%s://%s%s" % (self.scheme, domain, self.reverse_path)
         )
+        self.tz_api = (
+            "%s://%s%s" % (self.scheme, domain, self.timezone_path)
+        )
+        self.nearby_api = (
```
KostyaEsmukov: The naming style of these attributes should match the already existing one: `api_reverse` has `api` part in the beginning of the name instead of the end.

geopy/geocoders/geonames.py
```diff
@@ -125,6 +139,7 @@ def reverse(
             query,
             exactly_one=DEFAULT_SENTINEL,
             timeout=DEFAULT_SENTINEL,
+            find_nearby=False,
```
KostyaEsmukov: The definitely shouldn't be a boolean, because there're other reverse methods available in the service's API, such as `findNearbyPostalCodes`. It should be a choice-like string.

geopy/geocoders/geonames.py
```diff
@@ -171,13 +189,67 @@ def reverse(
             'lng': lng,
             'username': self.username
         }
-        url = "?".join((self.api_reverse, urlencode(params)))
+        url = "?".join((self.nearby_api, urlencode(params))) if find_nearby \
```
KostyaEsmukov: Just changing the url is not enough. These API endpoints accept different sets of parameters, so the arguments passed to this function should be validated for each API endpoint independently, and the incompatible ones should cause a `ValueError` being raised.

geopy/geocoders/geonames.py
```diff
@@ -171,13 +189,67 @@ def reverse(
...
+
+        response = self._call_geocoder(url, timeout=timeout)
+        try:
+            tz = timezone(response["timezoneId"])
+        except UnknownTimeZoneError:
+            raise GeocoderParseError(
+                "pytz could not parse the timezone identifier (%s) "
+                "returned by the service." % response["timeZoneId"]
+            )
+        except KeyError:
```
KostyaEsmukov: Too broad code sample is covered by this except clause. I.e. a KeyError raised by the `timezone` function is rather unexpected and shouldn't be processed by this clause.

KostyaEsmukov:
This is a great question regarding the other data returned by the timezone endpoint. I see the following options:
1. Return a custom wrapper instead of pytz.timezone, something similar to `Location`, which contains a `raw` attribute.
2. Move the actual implementation of the timezone method to another one, and make the pytz timezone construction in the current one, so the raw data could be retrieved by calling the second method.

I tend to incline towards the first option, but it would create inconsistency with the Google's timezone method.

I suggest you to just keep the pytz timezone for now. I'll reason about it before the release and will probably introduce the custom wrapper.

KostyaEsmukov:
@svalee I have just merged #332 which adds the aforementioned timezone wrapper. Could you please rebase your branch on top of the current master and use that wrapper in this PR?

KostyaEsmukov:
Hey @svalee, do you think you could work on this before I'd cut the next 1.18 release? If not I'd be happy to take this over so this feature could be finally published. :)

svalee:
@KostyaEsmukov hey, yeah, I am very sorry about this, when is 1.18 release?
I'll work on this today and tomorrow if I wouldn't be able to do it today, is this fine?

KostyaEsmukov:
> when is 1.18 release?

This PR is the main blocker, so as soon as it is merged.

> I'll work on this today and tomorrow if I wouldn't be able to do it today, is this fine?

Don't worry, it was just a friendly reminder from my side. This is fine. :)

test/geocoders/geonames.py
```diff
@@ -55,20 +55,62 @@ def test_query_urlencoding(self):
 
     def test_reverse(self):
```
KostyaEsmukov: This test contains too many checks. Could you please split it to multiple distinct test methods?

test/geocoders/geonames.py
```diff
@@ -55,20 +55,62 @@ def test_query_urlencoding(self):
...
+                **{
+                    "query": "40.75376406311989, -73.98489005863667",
+                    "exactly_one": True,
+                    "lang": 'ru'
+                }
+            )
+        self.assertEqual(loc.raw['adminName1'], 'Нью-Йорк')
+
+        with self.assertRaises(ValueError):
+            self._make_request(
```
KostyaEsmukov: Any reason to use the raw `self._make_request` instead of `self.reverse_run`? If no such, please change to `self.reverse_run`.

geopy/geocoders/geonames.py
```diff
@@ -163,9 +164,12 @@ def reverse(
             only, the value set during the geocoder's initialization.
 
         :param str feature_code: A GeoNames feature code
+            .. versionadded:: 1.18.0
 
         :param str lang: language of returned ``name`` element (the pseudo
             language code 'local' will return it in local language)
+            Full list of supported languages ISO639-2 can be found here:
+            http://www.loc.gov/standards/iso639-2/php/English_list.php
```
KostyaEsmukov: It's very unlikely that geonames does indeed support all countries from that list. I presume it should be `https://www.geonames.org/countries/` instead.

PR #329, opened 2018-10-21
 Adding Geolake geocoder

yholkamp:
This PR adds support for the Geolake geocoding service, it'd be great if this could make it in to the next release. Please let me know if you'd like me to make any additional changes.

### DIFF
--- a/geopy/geocoders/geolake.py
+++ b/geopy/geocoders/geolake.py
@@ -1,0 +1,163 @@
+from geopy.compat import urlencode
+from geopy.geocoders.base import DEFAULT_SENTINEL, Geocoder
+from geopy.location import Location
+from geopy.util import logger
+
+__all__ = ("Geolake", )
+
+
+class Geolake(Geocoder):
+    """Geocoder using the Geolake API.
+
+    Documentation at:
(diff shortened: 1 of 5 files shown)


geopy/geocoders/geolake.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+        self.api_key = api_key
+        self.domain = domain.strip('/')
+        self.api = '%s://%s%s' % (self.scheme, self.domain, self.api_path)
+
+    def geocode(
+            self,
+            query,
+            country_codes=None,
```
KostyaEsmukov: Could you please add a test case which uses this parameter?

yholkamp:
@KostyaEsmukov thanks for the rapid (and positive) review! Thanks for letting me know about the error on the sign up form and we're really sorry you had to run in to this. We've resolved this regression so this should not pose a problem again. I've also added the missed test scenario.

geopy/geocoders/geolake.py
yholkamp: Definitely, should be included in the latest commit.

PR #342, opened 2019-01-06
Fix opencage country parameter

### LINKED ISSUE #341: Opencage: country option in forward geocoding not working
It looks like opencage changed the request format. The new parameter is called `countrycode`instead of `country`. Also it expects now a ISO 3166-1 Alpha **2** instead of ISO 3166-1 Alpha **3**:
https://opencagedata.com/api#forward-opt

illing2005:
fixes #341

### DIFF
--- a/geopy/geocoders/opencage.py
+++ b/geopy/geocoders/opencage.py
@@ -113,11 +113,8 @@ class OpenCage(Geocoder):
                 This format is now deprecated in favor of a list/tuple
                 of a pair of geopy Points and will be removed in geopy 2.0.
 
-        :param str country: Provides the geocoder with a hint to the
-            country that the query resides in. This value will help the
-            geocoder but will not restrict the possible results to the
-            supplied country. The country code is a 2 character code as
-            defined by the ISO 3166-1 Alpha 2 standard.
+        :param str country:  Restricts the results to the specified country or countries. 
+            The country code is a 2 character code asdefined by the ISO 3166-1 Alpha 2 standard.
 
         :param bool exactly_one: Return one result or a list of results, if
             available.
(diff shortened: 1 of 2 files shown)


geopy/geocoders/opencage.py
```diff
@@ -116,8 +116,8 @@ def geocode(
         :param str country: Provides the geocoder with a hint to the
```
KostyaEsmukov: According to the docs at https://opencagedata.com/api#forward-opt, the `countrycode` param *restricts* the results to the specified country, however the current description there says that it just provides a *hint*. Could you please update this description?

illing2005: Good point! I changed that.

geopy/geocoders/opencage.py
```diff
@@ -113,11 +113,9 @@ def geocode(
                 This format is now deprecated in favor of a list/tuple
                 of a pair of geopy Points and will be removed in geopy 2.0.
 
-        :param str country: Provides the geocoder with a hint to the
-            country that the query resides in. This value will help the
-            geocoder but will not restrict the possible results to the
-            supplied country. The country code is a 3 character code as
-            defined by the ISO 3166-1 Alpha 3 standard.
+        :param str country: Restricts the results to the specified country or countries.
```
KostyaEsmukov: It is not clear how one could specify multiple countries. Should it be a comma- or a space-separated list, or should it be a python list of strings?)

I suppose this is not a single occurrence of such an ambiguity in geopy, so I'll address that for all geocoders later. To keep things "pythonic", it should accept a list of strings instead of a single comma-separated string (as specified in their docs at https://opencagedata.com/api#forward-opt).

illing2005:
Thanks for merging!
Do you already know when there will be a new release?

KostyaEsmukov:
TBH I didn't think of that yet. I'll try to get to it next week.

In the meantime you may want to consider installing from the master branch directly:

```
pip install --upgrade https://github.com/geopy/geopy/archive/master.zip
```

KostyaEsmukov:
I've just released `1.19.0` which contains this fix.

Apologies for the delay!

illing2005:
Great! Thanks for the info.

PR #345, opened 2019-02-08
Don't warn in GoogleV3 when using premier

mwhansen:
We're in the process of upgrading to the latest version of geopy and noticed we are getting warnings about the `api_key` when using the geocoder.  This PR removes the warning when the `client_id` and `secret_key` are passed in.


mwhansen:
Thanks!

PR #348, opened 2019-03-19
added place_id search parameter for geocode method

mstoncel:


geopy/geocoders/googlev3.py
```diff
@@ -159,6 +159,7 @@ def _format_components_param(components):
     def geocode(
             self,
             query=None,
+            place_id=None,
```
KostyaEsmukov: Could you please place this arg after `timeout`? Otherwise this would break the code which uses positional args, i.e. `g.geocode("london", False)`.

geopy/geocoders/googlev3.py
```diff
@@ -179,6 +180,11 @@ def geocode(
             .. versionchanged:: 1.14.0
                Now ``query`` is optional if ``components`` param is set.
 
+        :param str place_id: Address details are obtained using google
+                             place id.It is not used with the query or bounds
```
KostyaEsmukov: 1. There should be a space: `s/place id.It is/place id. It is/`
2. These extra lines should be indented by 4 spaces relative to the `:param` option, like in the other options there.

geopy/geocoders/googlev3.py
```diff
@@ -179,6 +180,11 @@ def geocode(
...
+                             parameter.
+                >>> g.geocode(place_id='ChIJOcfP0Iq2j4ARDrXUa7ZWs34')
```
KostyaEsmukov: I suppose there should be an extra newline between the code sample and the `:param ...:` option. https://github.com/geopy/geopy/blob/master/CONTRIBUTING.md#building-docs -- this guide explains how to build the docs, could you please do that and ensure that it is rendered correctly?

geopy/geocoders/googlev3.py
```diff
@@ -216,8 +222,18 @@ def geocode(
         params = {
             'sensor': str(sensor).lower()
         }
-        if query is None and not components:
+
+        if not any([query, components, place_id]):
             raise ValueError('Either `query` or `components` must be set.`')
```
KostyaEsmukov: The error message should include `place_id` as well.

geopy/geocoders/googlev3.py
```diff
@@ -216,8 +222,18 @@ def geocode(
...
+
+        if place_id and (bounds or query):
```
KostyaEsmukov: Looks like this should include the other 2 options as well.

I guess this would do it:
```
if sum(1 for value in (query, components, place_id) if value) != 1:
```

geopy/geocoders/googlev3.py
```diff
@@ -248,7 +264,6 @@ def geocode(
             url = self._get_signed_url(params)
         else:
             url = "?".join((self.api, urlencode(params)))
-
```
KostyaEsmukov: Could you please roll back this change? The new extra newline at the end the file too.

geopy/geocoders/googlev3.py
KostyaEsmukov: Oh, I didn't notice that it's `bounds` instead of `components`. Is that correct? If so, should the both conditions be `or`'ed: the yours and mine?

mstoncel: When the ``place_id`` parameter is used, it cannot be used with the ``bounds`` and / or ``query`` parameter. So I made this kind of limitation. Therefore, if the ``place_id`` is used, ``bounds`` and / or ``query`` will be not used. The ``components`` parameter can be used.

test/geocoders/googlev3.py
```diff
@@ -312,3 +312,50 @@ def test_geocode_bounds_invalid(self):
...
+                {"place_id": "ChIJOcfP0Iq2j4ARDrXUa7ZWs34",
+                 "query": "silicon valley"},
+                {}
+            )
+
+    def test_geocode_place_id_with_bounds(self):
+        with self.assertRaises(ValueError):
+            self.geocode_run(
+                {"place_id": "ChIJOcfP0Iq2j4ARDrXUa7ZWs34",
+                 "bounds": [50, -2, 55]},
```
KostyaEsmukov: These bounds are incorrect: the correct one should contain 4 coordinates, e.g. `[50, -2, 55, 2]`. Could you please fix that? Otherwise it's not clear whether the `ValueError` is caused by incorrect `bounds` value or by both `place_id` and `bounds` being specified.

geopy/geocoders/googlev3.py
```diff
@@ -179,6 +180,11 @@ def geocode(
             .. versionchanged:: 1.14.0
                Now ``query`` is optional if ``components`` param is set.
 
+        :param str place_id: Address information is obtained using google
```
KostyaEsmukov: The order of params in the docstring should match the order of arguments in the function signature. Could you please fix that? This `:param` block should be the last one.

geopy/geocoders/googlev3.py
KostyaEsmukov: OK, I see, thank you for the clarification.

PR #350, opened 2019-03-23
Deprecate `geocode`-specific args in `__init__`

KostyaEsmukov:
`__init__` args which are specific to the `geocode` methods have been deprecated in this PR.

The motivation is:
1. They raise questions like "do they affect other methods like `reverse`?"
2. They tend to not have a corresponding param in the `geocode` method, which forces to create new instances of geocoder for queries which need different values for such args.
3. Most of the `geocode` args still cannot be specified in the `__init__` methods.
4. They clutter the `__init__` method args which naturally should specify only the credentials and connectivity settings (well, except for the `format_string`).

As an alternative I'd suggest to use `functools.partial`:

```python
# Change this:
#   geocoder = GeocodeEarth("my_api_key", country_bias="US")
#   location = geocoder.geocode("new york")
# to this:

from functools import partial

geocoder = GeocodeEarth("my_api_key")
geocode = partial(geocoder.geocode, country_bias="US")
location = geocode("new york")
```

### DIFF
--- a/geopy/geocoders/geocodeearth.py
+++ b/geopy/geocoders/geocodeearth.py
@@ -41,7 +41,15 @@ class GeocodeEarth(Pelias):
                 This format is now deprecated in favor of a list/tuple
                 of a pair of geopy Points and will be removed in geopy 2.0.
 
+            .. deprecated:: 1.19.0
+                This argument will be removed in geopy 2.0.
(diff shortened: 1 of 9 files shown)


KostyaEsmukov:
Also related (Yandex geocoder): 2907bff

PR #358, opened 2019-05-09
Fixing the 'exactly_one' parameter usage in the mapbox geocoder

zeljkobekcic:
The mapbox geocoder returned exactly one argument even when setting `exactly_one=False` when geocoding. The problem was that the `_parse_json` method never recieved the said argument and used its default value `exactly_one=True`.


zeljkobekcic:
I have added a simple test which just checks if the result is a list. Do you think this is enough?

Issue #367, opened 2019-08-01
Add argument "namedetails" to Nominatim/OpenMapQuest


enriqueav:
Hi.

When geocoding using Nominatim, similar to the parameter already supported `addressdetails `, there is another one that is called `namedetails`:
https://nominatim.org/release-docs/develop/api/Search/#output-details

It basically adds all the alternative names of the place:
```
	"namedetails": {
		"name": "京都市",
		"name:cs": "Kjóto",
		"name:de": "Kyōto",
		...
		...
		"name:ja_kana": "きょうとし"
	}
```
I could work on this.

enriqueav:
I went ahead and create a PR #368

KostyaEsmukov:
Resolved by #368

PR #368, opened 2019-08-01
add namedetails argument to Nominatim, simplify country_codes logic

enriqueav:
Add namedetails argument to nominatim #367.

I needed to simplify the logic of the function because flake8 was giving error. 

```
$ flake8 geopy
geopy/geocoders/osm.py:210:5: C901 'Nominatim.geocode' is too complex (24) 
```

If you can think about another option to workaround this issue of complexity, please do let me know.

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -356,7 +356,9 @@ class Nominatim(Geocoder):
             params['bounded'] = 1
 
         if country_codes is None:
-            country_codes = self.country_bias or []
+            country_codes = self.country_bias
+        if not country_codes:
+            country_codes = []
         if isinstance(country_codes, string_compare):
             country_codes = [country_codes]
         if country_codes:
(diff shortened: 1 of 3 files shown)


geopy/geocoders/osm.py
```diff
@@ -351,9 +356,7 @@ def geocode(
             params['bounded'] = 1
 
         if country_codes is None:
-            country_codes = self.country_bias
-        if not country_codes:
-            country_codes = []
+            country_codes = self.country_bias or []
```
KostyaEsmukov: Could you please revert this change?

1. It changes behaviour: `country_codes=""` previously was coercing to `[]`, now it would be turning to `[""]`.

2. The previous option was easier to read: at first we fallback to `self.country_bias` if the `country_codes` was not set, then we clean up the value regardless where it came from. These are 2 separate steps.

3. This change is unrelated to the feature you're adding in this PR.

KostyaEsmukov:
I have just verified that the added test passes for OpenMapQuest and PickPoint as well, so this is good to go as soon as the `country_codes` change would be reverted.

geopy/geocoders/osm.py
enriqueav: Thanks for the feedback and the explanation.
I have reverted the change and just increased the max complexity.

enriqueav:
Thanks for the feedback and the explanation.
I have reverted the change and just increased the max complexity.

PR #372, opened 2019-09-20
Travis: use xenial dist globally; sudo is not used

mwtoews:
This PR brings a few modern features of Travis CI:

* As of [April 2019](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment), `xenial` is the default build environment. But specify this at the top so that it is clear.
* The `sudo` keyword [is deprecated](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration#timeline---its-happening-fast) (and probably does nothing)
* Put the tests for 3.7 and 3.8-dev in the regular matrix
* Update lint and rst tests to use Python 3.7, which is current (for another month or two, at least)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: python
-sudo: false
+dist: xenial
 
 python:
   - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
-  # 3.7 and 3.8 are defined separately in the jobs below.
-  # See: https://github.com/travis-ci/travis-ci/issues/9815
+  - "3.7"
+  - "3.8-dev"
   - "pypy"
   - "pypy3"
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,28 +36,17 @@ jobs:
...
     - stage: rst
       name: "RST (README.rst + docs) syntax check"
-      python: "3.6"
+      python: "3.7"
       before_script:
         - travis_retry pip install -e ".[dev-docs]"
       script:
         - make rst_check
 
-    # The `test` stage using the `python` matrix above is included implicitly.
```
KostyaEsmukov: Could you please return this comment back?

The rest of the diff looks good to me, thank you!

mwtoews: ah right, the comment is still valid and helpful

PR #378, opened 2019-11-18
Add language parameter to the Pelias geocoder

mir06:
Both methods `geocode` and `reverse` accept the `language` parameter which will be passed to the pelias-api as GET parameter `lang`.

This applies to `GeocodeEarth` as well.

### DIFF
--- a/geopy/geocoders/pelias.py
+++ b/geopy/geocoders/pelias.py
@@ -203,7 +203,7 @@ class Pelias(Geocoder):
 
         if language:
             params["lang"] = language
-            
+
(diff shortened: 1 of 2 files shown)


mir06:
@KostyaEsmukov I have corrected the lint errors. Now there is a problem with the yandex geocoder, which I did not touch.

This problem obviously appeared in the last three months as the last test of master worked (https://travis-ci.org/geopy/geopy/builds/588060571?utm_medium=notification&utm_source=github_status)

mir06:
Well, thank you very much for this great package!

PR #379, opened 2019-11-19
Added new api call (lookup) for the Nominatim geocoder

hyper750:
Added new api call (lookup) to be able to retrieve Location objects using the osm_id and osm_type.


hyper750:
> any chance to add it?

I will resolve the conflicts, but no one merged it in 4 years... :/

RaczeQ:
Don't know if it will change anything, but this change would be very welcomed 😄

PR #388, opened 2020-02-02
Support here apikeys

deeplook:


KostyaEsmukov:
Thank you! I'll merge this as-is, but will add some changes on top of master, such as:
- the new `apikey` argument breaks backwards compatibility, so it should have a default value `None` and be located after `app_id` and `app_code`
- tests should exercise both `app_id`+`app_code` and `apikey` options, not just one of them

But these are just little details, thanks again for providing a working fix!

deeplook:
Thanks!

PR #335, opened 2018-11-17
geopy 2.0

KostyaEsmukov:
This is a branch where geopy 2.0 development takes place. See #292 for the plan.

All new PRs should still be made against the `master` branch. I'll port them to this branch by myself.

I'll extensively do force-pushes to this branch, so please don't do anything on top of it.

PR #390, opened 2020-02-03
Fix collections.abc import for Python 3.9

hugovk:
The Python 3.9 build is failing:

```
geopy/point.py:264: in __eq__
    if not isinstance(other, collections.Iterable):
E   AttributeError: module 'collections' has no attribute 'Iterable'
```

https://travis-ci.org/geopy/geopy/jobs/645182137#L297

Python 3.8 warns about the removal:

```
  /home/travis/build/geopy/geopy/geopy/point.py:264: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```

https://travis-ci.org/geopy/geopy/jobs/645182134#L306

This change fixes the 3.9 build.

### DIFF
--- a/geopy/point.py
+++ b/geopy/point.py
@@ -3,7 +3,6 @@
 :class:`.Point` data structure.
 """
 
-import collections
 import re
 import warnings
 from itertools import islice
@@ -12,6 +11,14 @@ from math import fmod
 from geopy import units, util
 from geopy.compat import isfinite, string_compare
 from geopy.format import DEGREE, DOUBLE_PRIME, PRIME, format_degrees, format_distance
+
+try:
+    # Python 3.3+
+    from collections.abc import Iterable
+except ImportError:
+    # Python 2.7
+    from collections import Iterable
+
 
 POINT_PATTERN = re.compile(r"""
     .*?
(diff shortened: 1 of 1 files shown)


hugovk:
The PyPy3 failure is an unrelated timeout.

It passed on my fork: https://travis-ci.com/hugovk/geopy/jobs/283044372

KostyaEsmukov:
I'm sorry but I will have to close this, because geopy 1.x will not support Python 3.9, but geopy 2.0 branch already contains this fix, see #335

hugovk:
OK! How about removing the failing 3.9 testing from the 1.x/master branch?

KostyaEsmukov:
Thanks for understanding!

> How about removing the failing 3.9 testing from the 1.x/master branch?

Would that solve any problem? :) 

I think it's good to have it here to keep reminded that 1.x branch doesn't work with 3.9

hugovk:
It'd save some resources, free up some slots on the CI, save some electricity.

Seeing the failing build also suggests you do want eventually it to work with 3.9, but it's not yet ready, hence this PR :)

PR #397, opened 2020-02-18
Add MapTiler

chilfing:


sfkeller:
- [ ] Petr ( @klokan from Maptiler) OK?
- [ ] KostyaEsmukov OK?

geopy/geocoders/maptiler.py
```diff
@@ -0,0 +1,189 @@
...
+from geopy.util import logger
+
+__all__ = ("MapTiler", )
+
+
+class MapTiler(Geocoder):
+    """Geocoder using the MapTiler API.
+
+    Documentation at:
+        https://cloud.maptiler.com/geocoding/?_ga=2.134043661.283487034.1581321159-amp-OxJ6mv8GSFOKBYaTzxmOkQ
```
KostyaEsmukov: Could you please remove the `_ga` cookie?

Also, the link doesn't seem to be working without registration/authentication. Is there a publicly available doc which doesn't require authentication?

test/geocoders/maptiler.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        self.assertIn("チューリッヒ", location.address)
+
+    def test_geocode_raw(self):
+        result = self.geocode_run({"query": "New York"}, {})
+        self.assertTrue(isinstance(result.raw, dict))
+        self.assertEqual([-73.8784155, 40.6930727], result.raw['center'])
+        self.assertEqual("relation175905", result.raw['properties']['osm_id'])
+
+    def test_geocode_exactly_one_true(self):
```
KostyaEsmukov: The test is called `exactly_one_true`, but the code exercises `"exactly_one": False` case. Could you please rename the test to fix the ambiguity?

test/geocoders/maptiler.py
```diff
@@ -0,0 +1,97 @@
...
+            {"query": "Zürich", "exactly_one": True, "language": "ja",
+             "proximity": Point(47.3723, 8.5422)},
+            {"latitude": 47.3723, "longitude": 8.5422, "delta": 1},
+        )
+        self.assertIn("チューリッヒ", location.address)
+
+    def test_geocode_raw(self):
+        result = self.geocode_run({"query": "New York"}, {})
+        self.assertTrue(isinstance(result.raw, dict))
+        self.assertEqual([-73.8784155, 40.6930727], result.raw['center'])
```
KostyaEsmukov: It would compare floats without any epsilon, thus such test might fail randomly due to little fluctuations in floats.

Could you please change the assertion to `assertAlmostEqual` with a `delta`? There's an example of that, but here it could be simpler, in just 2 lines of `assertAlmostEqual` checks for lat and long: https://github.com/geopy/geopy/blob/61bbab8d041c819d15cb28249686220222e2a694/test/geocoders/util.py#L111-L126

geopy/geocoders/maptiler.py
chilfing: sadly I could not find any publicly available docs

klokan:
Very happy to see the implementation of the geocoder in geopy. Thanks!

Documentation is available at https://cloud.maptiler.com/geocoding/.
You need to create an account to use the service.

sfkeller:
@ KostyaEsmukov Did you see the review request?

KostyaEsmukov:
Whoops, didn't mean to close it.

Resolved conflicts and merged as ad31b19a16af8e92e6667fd082dafba43d4e3244. Thanks!

PR #398, opened 2020-03-04
googlev3: Add error_message to error checking

mrshu:
Hello everyone,

First of all, thanks a ton for putting `geopy` together!

While playing with the `geocoder` module and setting up GoogleV3 API, I ran into a few issues. The contents of `error_message` has helped me to debug it much faster and so I believe it would be good to make it part of the exception that it being raised.

I am not sure if I followed the standards of the `geopy` codebase, so please feel free to point that out and I'd be happy to refactor the PR to reflect it better.

Thanks!

---------------------------------

* GoogleV3 API response contains an `error_message` field when `status` is
  not `OK`. This field contains various complementary information that
  helps to debug the issue with the provided geocoder query.

  https://developers.google.com/maps/documentation/geocoding/intro#ErrorMessages

* This commit adds the content of the `error_message` field to the
  GeocoderQueryError exception in cases when `error_message` is provided
  in the API response and the exception is to be raised.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/geopy/geocoders/googlev3.py
+++ b/geopy/geocoders/googlev3.py
@@ -439,7 +439,7 @@ class GoogleV3(Geocoder):
     def _parse_json_timezone(self, response):
         status = response.get('status')
         if status != 'OK':
-            self._check_status(status)
+            self._check_status(status, response.get('error_message'))
 
         timezone_id = response.get("timeZoneId")
         if timezone_id is None:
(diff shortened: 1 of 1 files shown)


mrshu:
cc @KostyaEsmukov @jbouvier

mrshu:
@KostyaEsmukov @jbouvier This is just a friendly reminder that I'd be happy to address any comments you guys may have.

Thanks!

KostyaEsmukov:
e1c00c4 should fix the issue: the API's `error_message` will now be passed to the exception messages when it's present (which, I suppose, should almost always be true).

PR #403, opened 2020-04-10
Update link to Nominatim documentation

lonvia:
The documentation for Nominatim is no longer maintained in the OSM wiki. Point to the latest documentation on nominatim.org instead. That is the one that is valid for nominatim.openstreetmap.org.


lonvia:
Sure. I've amended the commit accordingly.

PR #406, opened 2020-04-19
Implement Nominatim reverse geocoder parameter "zoom"

quanterium:
Implemented the "zoom" parameter for the Nominatim reverse geocoder. This allows the user to select the level of detail for the address.

### DIFF
--- a/geopy/geocoders/osm.py
+++ b/geopy/geocoders/osm.py
@@ -412,7 +412,8 @@ class Nominatim(Geocoder):
             exactly_one=True,
             timeout=DEFAULT_SENTINEL,
             language=False,
-            addressdetails=True
+            addressdetails=True,
+            zoom=None
     ):
         """
         Return an address by location point.
@@ -443,6 +444,8 @@ class Nominatim(Geocoder):
 
             .. versionadded:: 1.14.0
 
+        :param str zoom: Level of detail required for the address.
+
         :rtype: ``None``, :class:`geopy.location.Location` or a list of them, if
             ``exactly_one=False``.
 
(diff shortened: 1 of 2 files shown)


geopy/geocoders/osm.py
```diff
@@ -443,6 +444,8 @@ def reverse(
 
             .. versionadded:: 1.14.0
 
+        :param str zoom: Level of detail required for the address.
```
KostyaEsmukov, line 447: If my understanding is correct, `zoom` would usually be a float or an integer rather than a string.

Maybe the type should be documented as union?

```
:param str zoom: Level of detail required for the address.
:type zoom: str, int, or float.
```

test/geocoders/nominatim.py
```diff
@@ -341,6 +341,18 @@ def test_namedetails(self):
...
+        query = "33.8120962, -117.9211682"
+        result_reverse = self.reverse_run(
+            {"query": query, "exactly_one": True, "zoom": "10"},
+            {},
+        )
+
+        self.assertEqual(
+            result_reverse.raw["display_name"],
+            "Anaheim, Orange County, California, United States of America"
+        )
```
KostyaEsmukov, line 354: Could you please add to this test the same query but without `zoom`, which would demonstrate that the `zoom` parameter actually made the response different?

geopy/geocoders/osm.py
quanterium: Done.

test/geocoders/nominatim.py
quanterium: Done.

Issue #430, opened 2020-07-14
HERE.com's new Geocoding & Search API


pratheekrebala:
Hi @KostyaEsmukov,

Just noticed that HERE.com is no longer maintaining the "Geocoding" API but are instead re-directing folks to use the "Geocoding and Search API". I haven't used both of them extensively but it looks like the new API better handles ambiguous addresses and supports providing component level filtering (e.g. city, county etc). 

Is there any interest in extending support to that API? I can make a PR if you are open to this. Both these APIs are in-compatible so it would either have to break the existing one or create a new Geocoder class (I prefer this approach).

https://developer.here.com/documentation/geocoder/dev_guide/topics/what-is.html

![Screenshot from 2020-07-14 11-37-55](https://user-images.githubusercontent.com/1174162/87445799-7d069000-c5c6-11ea-8d88-666b99d54993.png)

KostyaEsmukov:
I think it might be useful to add a new class for the new API (e.g. `HereV7`). So yeah, a PR would be very welcome!

If they share a lot, then it can be implemented similarly to `BaiduV3` which extends `Baidu`: https://github.com/geopy/geopy/blob/7e276b43fd66d603e8fd0c25313896ee324e1d49/geopy/geocoders/baidu.py#L275-L283

If they don't, it should extend the base geocoder class as the most of the other geocoders.

pratheekrebala:
I think they are just different enough that it might warrant a separate class. I wrote a small wrapper that I have been using internally - I will clean up and make a PR before EOD

KostyaEsmukov:
Fixed via #433

PR #433, opened 2020-07-21
Add support for HERE v7 geocoding service

pratheekrebala:
Addressing #430 

Still needs documentation and tests

### DIFF
--- a/geopy/geocoders/herev7.py
+++ b/geopy/geocoders/herev7.py
@@ -77,8 +77,6 @@ class HereV7(Geocoder):
 
         :param callable adapter_factory:
             See :attr:`geopy.geocoders.options.default_adapter_factory`.
-
-            .. versionadded:: 2.0
         """
         super().__init__(
             scheme=scheme,
@@ -100,13 +98,11 @@ class HereV7(Geocoder):
     def geocode(
         self,
         query,
-        components=None,
         *,
         bbox=None,
         exactly_one=True,
         maxresults=None,
         language=None,
-        additional_data=None,
         timeout=DEFAULT_SENTINEL
     ):
         """
--- a/geopy/geocoders/herev7.py
+++ b/geopy/geocoders/herev7.py
@@ -14,6 +14,7 @@ from geopy.exc import (
 from geopy.geocoders.base import DEFAULT_SENTINEL, Geocoder
 from geopy.location import Location
 from geopy.util import join_filter, logger
+from geopy.point import Point
 
 __all__ = ("HereV7", )
 
@@ -99,10 +100,13 @@ class HereV7(Geocoder):
         self,
         query,
         *,
+        components=None,
+        circle=None,
+        country=None,
         bbox=None,
+        language=None,
         exactly_one=True,
         maxresults=None,
-        language=None,
         timeout=DEFAULT_SENTINEL
     ):
         """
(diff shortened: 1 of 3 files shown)


geopy/geocoders/herev7.py
```diff
@@ -0,0 +1,286 @@
...
+)
+from geopy.geocoders.base import DEFAULT_SENTINEL, Geocoder
+from geopy.location import Location
+from geopy.util import join_filter, logger
+
+__all__ = ("HereV7", )
+
+
+class HereV7(Geocoder):
+    """Geocoder using the HERE Geocoder API.
```
KostyaEsmukov, line 22: Could you please describe the difference from the `Here` class, so it would be clear which one to use? (An also in `Here` as well). Example:
- https://github.com/geopy/geopy/blob/24b9226e1539a96ce379fad0b457fcfe37b3d6be/geopy/geocoders/baidu.py#L24-L26
- https://github.com/geopy/geopy/blob/24b9226e1539a96ce379fad0b457fcfe37b3d6be/geopy/geocoders/baidu.py#L276

geopy/geocoders/herev7.py
```diff
@@ -0,0 +1,286 @@
...
+        'country'
+    }
+
+    geocode_path = '/v1/geocode'
+    reverse_path = '/v1/revgeocode'
+
+    def __init__(
+            self,
+            *,
+            apiKey=None,
```
KostyaEsmukov, line 45: In Python arguments in public interfaces should adhere to snake_case. Could you please rename that to `apikey` or `api_key`?

geopy/geocoders/herev7.py
```diff
@@ -0,0 +1,286 @@
...
+            See :attr:`geopy.geocoders.options.default_user_agent`.
+
+        :type ssl_context: :class:`ssl.SSLContext`
+        :param ssl_context:
+            See :attr:`geopy.geocoders.options.default_ssl_context`.
+
+        :param callable adapter_factory:
+            See :attr:`geopy.geocoders.options.default_adapter_factory`.
+
+            .. versionadded:: 2.0
```
KostyaEsmukov, line 78: This `versionadded` directive is not needed for that param, because the whole HereV7 class would be added later than the already released 2.0.

geopy/geocoders/herev7.py
```diff
@@ -0,0 +1,286 @@
...
+        self.apiKey = apiKey
+        self.api = "%s://geocode.%s%s" % (self.scheme, domain, self.geocode_path)
+        self.reverse_api = (
+            "%s://revgeocode.%s%s" % (self.scheme, domain, self.reverse_path)
+        )
+
+    def geocode(
+        self,
+        query,
+        components=None,
```
KostyaEsmukov, line 100: `components` should be behind the asterisk as well, because the base `geocoder` method doesn't have that arg.

geopy/geocoders/herev7.py
```diff
@@ -0,0 +1,286 @@
...
+        *,
+        bbox=None,
+        exactly_one=True,
+        maxresults=None,
+        language=None,
+        additional_data=None,
```
KostyaEsmukov, line 106: Please remove that arg, see the last paragraph at https://github.com/geopy/geopy/blob/master/CONTRIBUTING.md#improving-a-geocoder

geopy/geocoders/herev7.py
pratheekrebala: Fixed in 078c382

geopy/geocoders/herev7.py
pratheekrebala: Fixed in 445fbac

geopy/geocoders/herev7.py
pratheekrebala: Fixed in 79abd5d

geopy/geocoders/herev7.py
pratheekrebala: Fixed in 3072157

geopy/geocoders/herev7.py
pratheekrebala: Fixed in 3072157. I added the following args instead:

- bbox
- circle
- country

pratheekrebala:
Thank you for your patience on this. I'm working on adding tests.

KostyaEsmukov (approved):
Thanks! I'm merging this, but will do some changes on top in master.

PR #434, opened 2020-08-03
Upgraded What3Words API to v3

saidtezel:

PR #444, opened 2020-09-09
What3Words API v3

saidtezel:
Added new class and associated tests for the What3Words API v3.

Supersedes #434


KostyaEsmukov (approved):
Thanks! Sorry for waiting for so long. Gonna make some changes on top of that (mainly -- collapse `what3words.py` and `what3wordsv3.py` to a single module and reuse tests via inheritance).

PR #447, opened 2020-09-30
Add USCensus geocoder

tylerlaws:
Add geocoder that uses the United States Census Bureau geocode API

### DIFF
--- a/geopy/geocoders/uscensus.py
+++ b/geopy/geocoders/uscensus.py
@@ -1,0 +1,189 @@
+from functools import partial
+from urllib.parse import urlencode
+
+from geopy.geocoders.base import DEFAULT_SENTINEL, Geocoder
+from geopy.location import Location
+from geopy.point import Point
+from geopy.util import logger
+
+__all__ = ("USCensus", )
+
+
+class USCensus(Geocoder):
+    """ Geocoder that uses the United States Census Bureau geocode API
+
+    Documentation at:
+        https://geocoding.geo.census.gov/geocoder
+    """
+
+    structured_query_params = {
+        'street',
+        'city',
+        'state',
+        'zip',
+    }
+
+    def __init__(
+        self,
+        *,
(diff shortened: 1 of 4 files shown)


tylerlaws:
Needs update to support Python 3.5

KostyaEsmukov:
Thanks!

I tried running your tests, and the service seems to be very slow for me:
```
geocoding.geo.census.gov  count: 12; min: 0.28s, median: 7.51s, max:13.93s, mean: 6.83s, total:81.94s
```

Is there a reason for that? I'm afraid that in that condition it would be quite painful to support :(

geopy/geocoders/uscensus.py
```diff
@@ -0,0 +1,189 @@
...
+        self,
+        query,
+        *,
+        dataset_type='Public_AR',
+        spatial_benchmark='Current',
+        geolookup=False,
+        geography_vintage='Current',
+        layers=None,
+        exactly_one=True,
+        timeout=16
```
KostyaEsmukov, line 82: Geopy geocoders should have `DEFAULT_SENTINEL` value as default for `timeout`. I have just merged additional tests in #480 which would check this property. So if you choose to continue working on getting this merged, please rebase on the fresh master branch and fix the test errors.

AVSISP:
> Thanks!
> 
> I tried running your tests, and the service seems to be very slow for me:
> 
> ```
> geocoding.geo.census.gov  count: 12; min: 0.28s, median: 7.51s, max:13.93s, mean: 6.83s, total:81.94s
> ```
> 
> Is there a reason for that? I'm afraid that in that condition it would be quite painful to support :(

It's because of where you are located. The US Census Tiger Geocoder isn't anycasted or load balanced. It's hosted on the USA east-coast. The slowness you're getting seems normal for someone accessing from a non-connected country (no direct routes to USA) such as Russia, some parts of Asia, or Oceania.

From a USA server located close to Virginia, it normally gets less than 150ms.

Issue #432, opened 2020-07-16
Failed to create Point instance from string: unknown format.


nseidl:
This example works:

geopy.point.Point('+037.33270,-122.00530')

This example does not work:

geopy.point.Point('+041.69438,+001.51461')

ValueError: Failed to create Point instance from string: unknown format.

The string in example that is not working is indeed valid GPS, so I believe that this is a bug.

KostyaEsmukov:
The following patch seems to fix the issue:

```
diff --git a/geopy/point.py b/geopy/point.py
index d6a5ee8..7c061e0 100644
--- a/geopy/point.py
+++ b/geopy/point.py
@@ -15,20 +15,20 @@ POINT_PATTERN = re.compile(r"""
     .*?
     (?P<latitude>
       (?P<latitude_direction_front>[NS])?[ ]*
-        (?P<latitude_degrees>-?%(FLOAT)s)(?:[%(DEGREE)sD\*\u00B0\s][ ]*
+        (?P<latitude_degrees>[+-]?%(FLOAT)s)(?:[%(DEGREE)sD\*\u00B0\s][ ]*
         (?:(?P<latitude_arcminutes>%(FLOAT)s)[%(PRIME)s'm][ ]*)?
         (?:(?P<latitude_arcseconds>%(FLOAT)s)[%(DOUBLE_PRIME)s"s][ ]*)?
         )?(?P<latitude_direction_back>[NS])?)
     %(SEP)s
     (?P<longitude>
       (?P<longitude_direction_front>[EW])?[ ]*
-      (?P<longitude_degrees>-?%(FLOAT)s)(?:[%(DEGREE)sD\*\u00B0\s][ ]*
+      (?P<longitude_degrees>[+-]?%(FLOAT)s)(?:[%(DEGREE)sD\*\u00B0\s][ ]*
       (?:(?P<longitude_arcminutes>%(FLOAT)s)[%(PRIME)s'm][ ]*)?
       (?:(?P<longitude_arcseconds>%(FLOAT)s)[%(DOUBLE_PRIME)s"s][ ]*)?
       )?(?P<longitude_direction_back>[EW])?)(?:
     %(SEP)s
       (?P<altitude>
-        (?P<altitude_distance>-?%(FLOAT)s)[ ]*
+        (?P<altitude_distance>[+-]?%(FLOAT)s)[ ]*
         (?P<altitude_units>km|m|mi|ft|nm|nmi)))?
     \s*$
 """ % {
```

[...]

PR #448, opened 2020-09-30
Add support for leading plus sign in POINT_PATTERN

azimjohn:
Closes https://github.com/geopy/geopy/issues/432


KostyaEsmukov (approved):
Thanks!

PR #450, opened 2020-10-08
Add additional information on which API products to enable for Google.

the-neil:


KostyaEsmukov:
I'm sorry but I'm going to have to decline this change.

The attention block tells only about the deprecation of api-key-less access, and it's been 2 years since that change so that block should rather be removed completely than be updated.

The required API names are already mentioned in the api_key doc: [geopy.geocoders.GoogleV3.\_\_init\_\_](https://geopy.readthedocs.io/en/stable/#geopy.geocoders.GoogleV3.__init__)

Issue #441, opened 2020-08-22
Add geocodeapi?


matkoniecz:
It appears to exist and offer a free level https://geocodeapi.io/#pricingSection

PR #453, opened 2020-11-07
Add geocodeapi

dmitriiweb:
#441


geopy/geocoders/geocodeapi.py
```diff
@@ -0,0 +1,235 @@
...
+            address, street, country, region, locality
+
+        """
+        params = self._get_params_reverse(query, size=size, layers=layers)
+        url = '{}?{}'.format(self.api_reverse, params)
+        callback = partial(self._parse_json, exactly_one=exactly_one)
+        return self._call_geocoder(url, callback, timeout=timeout, headers=self.headers)
+
+    def _get_params_geocode(self, query, **kwargs):
+        main_param = '?text={}'.format(query)
```
KostyaEsmukov, line 179: Passing user input without escaping is unsafe. Please remove that class-level `_param_templates` dict and switch to `urllib.parse.urlencode`. See other geocoders for example.

test/geocoders/geocodeapi.py
```diff
@@ -0,0 +1,74 @@
+import pytest
+
+from geopy import exc
+from geopy.geocoders import GeocodeAPI
+from geopy.point import Point
+
+from test.geocoders.util import BaseTestGeocoder, env
+
+
+@pytest.mark.skipif(
```
KostyaEsmukov, line 10: `skipif` marks are not used anymore, this should be removed.

test/geocoders/geocodeapi.py
```diff
@@ -0,0 +1,74 @@
...
+    not bool(env.get('GEOCODEAPI_KEY')),
+    reason="No GEOCODEAPI_KEY env variable set"
+)
+class TestGeocodeAPI(BaseTestGeocoder):
+    @classmethod
+    def make_geocoder(cls, **kwargs):
+        return GeocodeAPI(env.get('GEOCODEAPI_KEY'), **kwargs)
```
KostyaEsmukov, line 17: If `GEOCODEAPI_KEY` is mandatory for running these tests, it must be retrieved via `__getitem__` rather than with `get`: `env['GEOCODEAPI_KEY']`

KostyaEsmukov (changes requested):
Thanks for the contribution! I left some comments that should be addressed before this could be considered for merging.

dmitriiweb:
Hi @KostyaEsmukov ,
Could you check the last update - I can't find a reason for errors

PR #454, opened 2020-11-12
improve Nominatim.search docstring

kannes:
- `view_box` -> `viewbox` plus proper markup
- Specify what `viewbox` actually does


kannes:
Failed tests are unrelated to my changes from what I am seeing, only Photon failed, which I did not touch in any way.

KostyaEsmukov (approved):
Thanks! This is much better.

PR #464, opened 2021-01-15
OpenCage provider: use no_annotations=1 as returned data is not used

mtmail:
OpenCage results include https://opencagedata.com/api#annotations data by default. As this data isn't put into the Location response object it can be skipped in the request.

### DIFF
--- a/geopy/geocoders/opencage.py
+++ b/geopy/geocoders/opencage.py
@@ -123,6 +123,7 @@ class OpenCage(Geocoder):
         params = {
             'key': self.api_key,
             'q': query,
+            'no_annotations': 1,
         }
         if bounds:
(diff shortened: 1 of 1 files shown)


KostyaEsmukov:
I'm afraid this might count as a breaking change, because it would affect the contents of the `location.raw` value.

Is the benefit from disabling these by default really prominent?

KostyaEsmukov:
Well, it seems to reduce mean request time in tests by 20%, so I guess that's something.

I'm merging this, but will add a new method param to switch annotations and will make it on by default, in order to not break backwards compatibility.

KostyaEsmukov (approved):
Thanks!

mtmail:
Sounds good. Thanks a lot.

Issue #465, opened 2021-01-26
Add geocod.io


duckontheweb:
They have a pay-as-you-go option that allows you to start at no cost without needing to provide payment info: https://www.geocod.io/pricing/.

I've started creating a new geocoder class for this and can open a PR when it's ready.

KostyaEsmukov:
Merged in #468

PR #468, opened 2021-02-02
Add geocodio

duckontheweb:
Adds geocoder class for [Geocod.io API](https://www.geocod.io/). 

The API returns a `422` status code for invalid queries. This PR handles these using the `_geocoder_exception_handler` method, but it might be reasonable to map this code to a `GeocoderQueryError` in [`geopy.geocoders.base.ERROR_CODE_MAP`](https://github.com/geopy/geopy/blob/master/geopy/geocoders/base.py#L193-L206).

#465


duckontheweb:
@KostyaEsmukov Any feedback on this PR or thoughts on when it might be merged? Thanks!

geopy/geocoders/geocodio.py
```diff
@@ -0,0 +1,243 @@
...
+        super().__init__(
+            scheme=scheme,
+            timeout=timeout,
+            proxies=proxies,
+            user_agent=user_agent,
+            ssl_context=ssl_context,
+            adapter_factory=adapter_factory,
+        )
+        if api_key is None:
+            raise ConfigurationError('Must provide an api_key.')
```
KostyaEsmukov, line 71: api_key is the only authentication method here, so the `api_key` parameter should just be required (i.e. don't have any default), then this check would become redundant. Example: https://github.com/geopy/geopy/blob/2.1.0/geopy/geocoders/what3words.py#L29

geopy/geocoders/geocodio.py
```diff
@@ -0,0 +1,243 @@
...
+            ``exactly_one=False``.
+        """
+        if query is not None and \
+                any(p is not None for p in (city, state, postal_code, country)):
+            raise GeocoderQueryError('Address component must not be provided if '
+                                     'query argument is used.')
+        if street is not None and \
+                not any(p is not None for p in (city, state, postal_code)):
+            raise GeocoderQueryError('If street is provided must also provide city, '
+                                     'state, and/or postal_code.')
```
KostyaEsmukov, line 133: Is there a reason why these conditions must be checked on the client-side?

(I'm okay with keeping these unless they might change in future).

geopy/geocoders/geocodio.py
```diff
@@ -0,0 +1,243 @@
...
+
+        api = '%s://%s%s' % (self.scheme, self.domain, self.reverse_path)
+
+        url = "?".join((api, urlencode(params)))
+
+        logger.debug("%s.reverse: %s", self.__class__.__name__, url)
+        callback = partial(self._parse_json, exactly_one=exactly_one)
+        return self._call_geocoder(url, callback, timeout=timeout)
+
+    @staticmethod
```
KostyaEsmukov, line 199: Please make don't use staticmethod here. Those two are perfectly normal instance methods.

test/geocoders/geocodio.py
```diff
@@ -0,0 +1,79 @@
...
+        api_key = kwargs.pop('api_key') if 'api_key' in kwargs else env['GEOCODIO_KEY']
+        return Geocodio(api_key=api_key, **kwargs)
+
+    async def test_user_agent_custom(self):
+        geocoder = self.make_geocoder(user_agent='my_user_agent/1.0')
+        assert geocoder.headers['User-Agent'] == 'my_user_agent/1.0'
+
+    async def test_error_with_no_api_key(self):
+        with pytest.raises(exc.ConfigurationError):
+            self.make_geocoder(api_key=None)
```
KostyaEsmukov, line 22: You can just create an instance here:
```
    async def test_error_with_no_api_key(self):
        with pytest.raises(exc.ConfigurationError):
            Geocodio(api_key=None)
```

Then it won't be necessary to support retrieving the api_key from kwargs in `make_geocoder`

KostyaEsmukov (commented):
Thanks! Sorry for the late response.

Overall looks good, I left a couple of minor comments.

> The API returns a `422` status code for invalid queries. This PR handles these using the `_geocoder_exception_handler` method, but it might be reasonable to map this code to a `GeocoderQueryError` in [`geopy.geocoders.base.ERROR_CODE_MAP`](https://github.com/geopy/geopy/blob/master/geopy/geocoders/base.py#L193-L206).

I guess it won't break anything if we move the 422 error to the ERROR_CODE_MAP. However, keeping that in the geocoder's own `_geocoder_exception_handler` method has a benefit of being more explicit, and it will protect against possibility of future refactorings breaking that error processing for Geocodio.

Pick whatever you like the most, I'm okay with both options.

geopy/geocoders/geocodio.py
duckontheweb: Fixed in dbc5bf377cf3d6c42b211447eb31013e6a1115dc

geopy/geocoders/geocodio.py
duckontheweb: I think the only argument would be to save making an invalid request, which might count against rate limits or account request limits.

test/geocoders/geocodio.py
duckontheweb: Fixed in 24b45cb265fb267f640f23b6a83baa684c57b5c2

geopy/geocoders/geocodio.py
duckontheweb: Fixed in 5deb061124a933cc9b2d70c6db65e58a8bdf9b7b

duckontheweb:
> > The API returns a `422` status code for invalid queries. This PR handles these using the `_geocoder_exception_handler` method, but it might be reasonable to map this code to a `GeocoderQueryError` in [`geopy.geocoders.base.ERROR_CODE_MAP`](https://github.com/geopy/geopy/blob/master/geopy/geocoders/base.py#L193-L206).
> 
> I guess it won't break anything if we move the 422 error to the ERROR_CODE_MAP. However, keeping that in the geocoder's own `_geocoder_exception_handler` method has a benefit of being more explicit, and it will protect against possibility of future refactorings breaking that error processing for Geocodio.
> 
> Pick whatever you like the most, I'm okay with both options.

Okay, I'm a fan of being explicit, so I'll just stick with handling it in the `_geocodoer_exception_handler`.

KostyaEsmukov (approved):
Looks good, thank you!

PR #476, opened 2021-03-12
be more explicit in lat lon ordering

matkoniecz:
Not everyone knows where this cities are located and lat lon / lon lat ordering is a bane of GIS software

At least of my software


KostyaEsmukov (approved):
Thanks!

PR #478, opened 2021-03-22
Add multifield query for the ArcGIS geocoder

PaulQbFeng:
I kept the structure similar to some other geocoders like Here, Bing


KostyaEsmukov (changes requested):
Thanks! Could you please add some tests for your feature?

PR #481, opened 2021-04-01
Update Photon domain

yrafalin:
Update Photon domain to correspond to the new one https://photon.komoot.io/


KostyaEsmukov (approved):
Thanks!

PR #493, opened 2021-09-15
Implement basic geocode functionality for Amazon Location Services.

kevinglasson:
Haven't written the tests yet, I'm not sure how to go about it as it requires and AWS account (I believe, still to investigate a bit more).

Happy for some feedback and further direction, especially want some feedback on putting optional dependencies behind a feature flag.

Changes:
- Added `AmazonLocationServices` as a new `Geocoder`.
- Put the `AmazonLocationServices` `Geocoder` behind a 'feature flag' as it
  pulls in `boto3` as a dependency. This involves a bit of lazy importing
hackery to make sure the library functions as normal for everyone who **does
not** care about this particular `Geocoder`.

### DIFF
--- a/geopy/geocoders/__init__.py
+++ b/geopy/geocoders/__init__.py
@@ -187,6 +187,7 @@ __all__ = (
     # Also don't forget to pull up the list of geocoders
     # in the docs: docs/index.rst
     "AlgoliaPlaces",
+    "AmazonLocationServices",
     "ArcGIS",
     "AzureMaps",
     "Baidu",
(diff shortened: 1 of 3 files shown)


KostyaEsmukov:
boto3 is a quite large dependency, and pulling it just for a single geocoder, which many people won't even use, doesn't sound appropriate.

And it's not just about the size. geopy has it's own way of making http requests, which respects async/sync mode, ssl_context and proxy settings (see https://geopy.readthedocs.io/en/stable/#module-geopy.adapters). boto3 doesn't allow to use that.

I'm going to decline this PR for that reason. If boto3 is absolutely required, I'd suggest to keep the implementation of such a geocoder outside of geopy. Otherwise an implementation based on the geopy adapters using the raw HTTP API (e.g. https://docs.aws.amazon.com/location/latest/APIReference/API_SearchPlaceIndexForText.html) might be suitable for merging into geopy.

Thanks for the attempt and I'm sorry for such a late feedback!

kevinglasson:
No problems, that is totally fair - thank you.

PR #502, opened 2021-10-28
Allowing different mapbox endpoints.

dennisstritzke:
The mapbox api provides different endpoints that are
differentiated by the url path. Adding an endpoint
parameter that is used to template the api path and
providing a default parameter to preserve the current
behaviour. Doing so allows us to create a mapbox
geocoder and call the mapbox.places-permanent endpoint.

### DIFF
--- a/geopy/geocoders/mapbox.py
+++ b/geopy/geocoders/mapbox.py
@@ -16,7 +16,7 @@ class MapBox(Geocoder):
         https://www.mapbox.com/api-documentation/
     """
 
-    api_path = '/geocoding/v5/mapbox.places/%(query)s.json/'
+    api_path = '/geocoding/v5/%(endpoint)s/%(query)s.json/'
 
     def __init__(
             self,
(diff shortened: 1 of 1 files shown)


KostyaEsmukov:
I'm worried that the endpoints might differ in the set of parameters/filters they accept, the documentation (https://docs.mapbox.com/api/search/geocoding/#endpoints) seems to be formulated vaguely in this regard (at least as I see it). So introducing a separate geocoder class seems to be a more future-proof way of adding support for this.

For the time being I'd suggest to use this endpoint with something similar to this:

```py
import geopy.geocoders

class MapBoxPermanent(geopy.geocoders.MapBox):
    api_path = '/geocoding/v5/mapbox.places-permanent/%(query)s.json/'

geolocator = MapBoxPermanent(api_key='...')
...
```

> To enable the endpoint you have to contact MapBox sales. That took about 1h in my case.

According to their pricing page (https://www.mapbox.com/pricing#permanent-geocoding-api) this API doesn't provide a free tier, which makes it not applicable here.

I'm going to go ahead and close this, but feel free to open a new PR with addition of a new class if you feel like it should be present in the library.

PR #520, opened 2022-05-15
allow geographiclib 2.x

dimbleby:
surprisingly hard to track down an upstream changelog, https://geographiclib.sourceforge.io/html/C/changes.html suggests that the API is basically unchanged despite the major version bump.


KostyaEsmukov:
Thanks!

PR #606, opened 2026-05-11
Fix test_reverse_near_zero

atupone:
Test fails as returned string is in Arabic. This change will get the result in English so it succeeds


Ruiww:
感谢您的来信，我会尽快查阅并回复！

KostyaEsmukov (approved):

PR #611, opened 2026-07-12
Move project metadata to pyproject.toml, use dependency groups

mwtoews:
This modernises a few project aspects:

- Move project metadata from setup.cfg to pyproject.toml ([PEP 621](https://peps.python.org/pep-0621/))
  - Most of this was a 1:1 conversion, with a few exceptions
  - Remove the licence classifier, which is deprecated ([PEP 639](https://peps.python.org/pep-0639/))
  - Change requires-python from >=3.8 to >=3.8.1 to solve a discrepancy with flake8 (found via `uv sync`)
  - Change "find" strategy used by setuptools to use "include" rather than "exclude", which I find more robust
- Use dependency groups ([PEP 735](https://peps.python.org/pep-0735/)) for internal development rather than optional dependencies
  - Firstly, keep a few useful and documented "extras" which are `geopy[aiohttp,requests,timezone]`
  - Use the "dev" group to install everything, but also add tox and geopy with all extras; note [`uv sync`](https://docs.astral.sh/uv/concepts/projects/sync/#syncing-development-dependencies) includes "dev" by default, otherwise it can be installed via `pip install --group dev`
  - Move extras dev-lint, dev-test and dev-docs to dependency groups lint, test and docs, as is the convention
  - Use [`dependency_groups`](https://tox.wiki/en/latest/reference/config.html#dependency_groups) feature with tox>=4.22


KostyaEsmukov:
Sorry, it's not the time for such migration yet.

This requires recent setuptools which not all downstream consumers may have.

mwtoews:
Why is this not the time? Python projects should generally align along the [Python Packaging User Guide](https://packaging.python.org/en/latest/), using community-accepted [Python Enhancement Proposals (PEPs)](https://peps.python.org/). That's the point of this PR.

Setuptools has had support for all of these features for at least 4 years, [v61.0.0](https://setuptools.pypa.io/en/latest/history.html#v61-0-0) (24 Mar 2022). See [Configuring setuptools using pyproject.toml files](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). What incompatibilities do you see, and what downstream consumers are using setuptools from over 4 years ago??

PR #601, opened 2026-04-02
Fix incorrect type hints for timeout and language params in Nominatim and Photon

nuglifeleoji:
## Problem (fixes #590, #593, #594)

Static type checkers (Pylance, mypy) report spurious errors for standard usage of `geocode()` and `reverse()` in the Nominatim and Photon geocoders.

**Issue 1 – `timeout` parameter (#590)**

```python
geolocator.geocode("London", timeout=5)
# Pylance: Argument of type "Literal[5]" cannot be assigned to parameter
#          "timeout" of type "object" in function "geocode"
```

`DEFAULT_SENTINEL` is created via `type('object', (object,), ...)()`, giving it the class name `'object'`. Pylance treats this as `geopy.geocoders.base.object` (not `builtins.object`), so any integer or float is seen as incompatible.

**Issue 2 – `language` parameter (#593, #594)**

```python
geolocator.geocode("London", language="de")
# Pylance: Argument of type "Literal['de']" cannot be assigned to
#          parameter "language" of type "bool" in function "geocode"
```

`language=False` causes Pylance to infer the parameter type as `bool`.

## Fix

Add explicit type annotations to `geocode()` and `reverse()` in **Nominatim** and **Photon**:

- `timeout: Optional[Union[int, float]] = DEFAULT_SENTINEL`
- `language: Optional[str] = None`  (default changed from `False` to `None`;
  both are falsy so all `if language:` checks are unaffected)

The fix uses `typing.Optional`/`typing.Union` for Python 3.7 compatibility.

Made with [Cursor](https://cursor.com)

Issue #590, opened 2025-07-24
Bug: Incorrect type hint for timeout parameter in geolocator.geocode


obiwan111:
When using static type checkers like Pylance calling: 
geo = geolocator.geocode("London", timeout=5)

raises following error: 
Argument of type "Literal[5]" cannot be assigned to parameter "timeout" of type "object" in function "geocode"
  "builtins.object" is not assignable to "geopy.geocoders.base.object"

Erhan1706:
I am also getting the incorrect type hint for the "language" parameter with pylance. This code: 
location = self.geolocator.geocode(city, language="en", timeout=None) 

Raises this warning: 
Argument of type "Literal['en']" cannot be assigned to parameter "language" of type "bool" in function "geocode"
  "Literal['en']" is not assignable to "bool"

So I guess most of the parameters have to be double-checked for this problem

PR #616, opened 2026-09-04
Annotate timeout parameter to stop spurious static-typing errors

### LINKED ISSUE #604: Pylance reports false argument type errors for DEFAULT_SENTINEL-backed parameters
`Nominatim` and other geocoders use `DEFAULT_SENTINEL` as the default value for parameters such as `timeout`. Because `DEFAULT_SENTINEL` currently has no broad static type annotation, source-based type checkers can infer those parameters as accepting only the sentinel object type.

Minimal repro:

```python
from geopy.geocoders import Nominatim

geolocator = Nominatim(user_agent="city_map_poster", timeout=10)
```

With Pyright/Pylance this produces a false positive similar to:

```text
Argument of type "Literal[10]" cannot be assigned to parameter "timeout" of type "object" in function "__init__"
```

Runtime behavior and the documentation both allow numeric `timeout` values, and the code runs correctly. The issue is only in static inference from the sentinel default.

Expected behavior: source-based type checkers should not infer `DEFAULT_SENTINEL`-backed parameters as sentinel-only parameters, so documented calls like `timeout=10` should not be flagged.

afonsojanu:
Fixes #604 (and covers the same underlying cause as #590/#593/#594, already being addressed for the geocode()/reverse() methods in #601).

Nominatim() and PickPoint() default their timeout keyword to DEFAULT_SENTINEL, an instance of an ad hoc anonymous class created via `type('object', (object,), ...)`. With no explicit annotation on the parameter, Pyright infers its type purely from that default value, so a perfectly normal call like:

```python
Nominatim(user_agent="my_app", timeout=10)
```

gets flagged with something like:

```
Argument of type "Literal[10]" cannot be assigned to parameter "timeout" of type "object" in function "__init__"
```

even though it works fine at runtime and is exactly what the docs tell you to do. The base `Geocoder.__init__` that every geocoder inherits from has the identical issue, so I annotated `timeout` there too.

PickPoint's `__init__` already carries a comment saying its signature has to stay in sync with Nominatim's, so I updated both together.

Verified with pyright directly: the repro above type-checks clean with this change, and reverting it brings the error straight back. Ran the non-network parts of the test suite (`--skip-tests-requiring-internet`) for the three touched modules, 43 passed, nothing broken.

One thing worth flagging: most of the other geocoder classes (arcgis, bing, google, mapbox, etc.) have the same DEFAULT_SENTINEL pattern on their own timeout/proxies/ssl_context parameters, so they'd hit the identical Pyright complaint. Fixing all of those together felt like a bigger, separate pass across two dozen-plus files, so I kept this PR scoped to what #604 actually reported.


Ruiww:
感谢您的来信，我会尽快查阅并回复！

